PR #1, opened 2020-07-29
Update debug.json

typebrook:
Add hint for API key

### DIFF
--- a/mbgl/debug.json
+++ b/mbgl/debug.json
@@ -4,7 +4,7 @@
   "sources": {
     "protomaps": {
       "type": "vector",
-      "tiles":["https://api.protomaps.com/tiles/v1/{z}/{x}/{y}.pbf"],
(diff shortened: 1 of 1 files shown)


bdon:
Sorry, I didn't get a notification for this PR for some reason...

I would recommend adding an API key via the "transformRequest" option in your JS code:

```
this.map = new mapboxgl.Map({
      container: 'map',
      style: 'http://localhost/mbgl/debug.json',
      transformRequest: u => {
        return {url: u + "?key=abcdef1234"}
      },
    })
```

I think that is preferable to embedding the key directly in the Style JSON, that way multiple applications can use different keys but share the same style file. What do you think?

typebrook:
This works only on Mapbox GL JS, you also need to consider usage on tangram or Mapbox GL native.

bdon:
I have added a preprocessor to avoid duplicating all of the Tangram and MapboxGL styles, this also takes a `-key` argument in case you want to version your customizations separately. I have also added instructions on setting key like above, or via code in the README

PR #22, opened 2023-03-06
fix: Places.java parseIntOrNull

tyrauber:
sf.getString(“population”) may be null. Must check first if present before attempting to parse.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -32,12 +32,11 @@ public class Places implements ForwardingProfile.FeatureProcessor, ForwardingPro
             } else if (sf.hasTag("place","city")){
                 feat.setAttr("pmap:kind", "city")
                         .setZoomRange(4, 15);
-                int population = Parse.parseIntOrNull(sf.getString("population"));
-                if (population > 200000) {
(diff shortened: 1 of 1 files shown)


tyrauber:
@bdon, this resolves the error when importing the planet.osm in planetiler using the pmtiles build profile.

I am curious though what the use of pmap:rank is, and why there are only two options (1&2) for cities with a population greater than and less than 200,000.    I assume this is used to determine which cities to render at what zoom level? I am surprised there are only two options?

Wouldn't it make more sense to set a population attribute (or perhaps population divisible by 100,000) and have the render logic happen client side in the layer javascript?

bdon:
`population` is not a reliable measurement in OSM, and anyways doesn't directly map to our cartographic goals; it's a temporary proxy measurement for now until we add something like https://qrank.wmcloud.org . 

Storing a raw measure in the tiles is possible, but another goal of keeping both a style + tiles in this one repository is to have some reasonable design shared among them - the style in this repo is "privileged" or opinionated. The existing design is to have ranks 1 and 2 correspond to different font weights or sizes in the style. Of course, the purpose of making this open source is you are free to modify the creation process if your cartographic goals are different (but how this exposed isn't decided yet)

bdon:
Thanks for the fix!

PR #35, opened 2023-04-28
Filter old tracks

HeikoGr:
according to: https://wiki.openstreetmap.org/wiki/Lifecycle_prefix

This should remove some old, unused and removed railways/highways.

https://taginfo.openstreetmap.org/tags/razed%3Arailway
especially railways looked weird in my area without this filter

Todo: update docs (I didn't find them in this repo)


bdon:
Thanks for the enhancement!

PR #38, opened 2023-06-01
tileset: fix population parsing of commas.

bdon:
@nvkelso this ignores OSM population tag values that have commas in them (though we should either fix them upstream or handle those correctly)

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -1,5 +1,7 @@
 package com.protomaps.basemap.layers;
 
+import static com.onthegomap.planetiler.util.Parse.parseIntOrNull;
+
 import com.onthegomap.planetiler.FeatureCollector;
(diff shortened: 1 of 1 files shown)


nvkelso (approved):
LGTM

Longer term there's a lot of junk in OSM tag values that can be sanitized.

Looks like there are other useful built-in functions for sanitizing, like:

- [meters](https://github.com/onthegomap/planetiler/blob/390bd3389f8efe0ab0bb454793a4ffff88fd8941/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Parse.java#L168-L211) - not sure why this isn't called `parseMeters`?
- [parseRoundInt](https://github.com/onthegomap/planetiler/blob/390bd3389f8efe0ab0bb454793a4ffff88fd8941/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Parse.java#L80-L91) which uses regex in TO_ROUND_INT_SUBSTRING_PATTERN.

So any new logic should be compatible with an upstream merge.

Issue #42, opened 2023-06-12
Road casings ordering leads to "bridge" like rendering


Edefritz:
Right now all road casings are rendered just before the road type they belong to. This leads to some weird effects at intersections between different road types (minor-medium/medium-major ...), where it appears that the higher class road passes above the lower class road, instead of being just a regular intersection.

<img width="700" alt="Screenshot 2023-06-12 at 11 13 25" src="https://github.com/protomaps/basemaps/assets/1841670/84e6794b-e124-4321-ba28-96e34cd623ec">
<img width="700" alt="Screenshot 2023-06-12 at 11 13 38" src="https://github.com/protomaps/basemaps/assets/1841670/a1c62745-87b6-4de3-9ba9-e9079c675f8d">

I'm not a cartography expert, but I think road casings should all be rendered before (below) any of the roads. I will propose a PR later today, if that's ok.

nvkelso:
+1

wipfli:
Thanks for this contribution, @Edefritz! I agree that in general it is better to do first all casings, then all line fills for roads.

Nevertheless, I would like to mention that there are also downsides to first render all casings, and then all line fills:

1) Roads of lower priority which merge perpendicularly into a road with higher priority will "bite out" a section of the casing of the high priority road which breaks the visual continuity of the high priority road. This effect is in particular visible at low zooms.
2) A low priority road parallel and close to a high priority road will delete the casing of the high priority road. It looks effectively like the high priority road has no casing. Again, this effect is strongest at low zooms.

Here is an example from the Wikipedia basemap which has these problems (https://maps.wikimedia.org/#15/47.2109/7.9993):

![Screenshot_20230625_214632](https://github.com/protomaps/basemaps/assets/53421382/11beb957-03c6-4045-a0e6-6c0ab4735d6d)

Google Maps solves this issue by using casing-fill-casing-fill order at low zooms, while switching to casing-casing-fill-fill at higher zooms. Here is an example:

At z13: 

https://www.google.com/maps/@47.2076993,8.0005265,13z?entry=ttu

![image](https://github.com/protomaps/basemaps/assets/53421382/417a9de3-b92a-4cde-9a3d-f603e672a8ac)

At z14:

https://www.google.com/maps/@47.2076993,8.0005265,14z?entry=ttu

![image](https://github.com/protomaps/basemaps/assets/53421382/a8f17c59-615c-487c-86e0-6fe77a5d86df)


### Side Note: Bridges

[...]

PR #43, opened 2023-06-12
Place road casings below all roads

Edefritz:
Prevents "bridge" like rendering effect of regular intersections.

See https://github.com/protomaps/basemaps/issues/42

Now I'm not 100% sure this is the correct approach. Maybe there is something else I didn't consider so someone should definitely double check for other cases. 
But simply moving the casings for all four road types above the first rendered road type seems to address the issue and produces the desired outcome.

OSM
Screenshot 2023-06-12 at 11 13 38
Before PR
Screenshot 2023-06-12 at 11 13 25
After PR
Screenshot 2023-06-12 at 11 12 46


nvkelso:
+1 to this general approach. I'll followup with a code review soon.

Edefritz:
Thanks! :) I just noticed the diffs might look a bit confusing. The only thing I did was moving the road casing layers up in the rendering order. Git makes it seems as if I changed individual properties of some existing layers.

bdon:
Thanks! Looks like we need to do the same for tunnels and bridges (all casings for all classes instead before strokes)

Edefritz:
Thanks again for looking into this. I guess now that the changes were merged, it would also make sense to increment the version of the [npm package](https://www.npmjs.com/package/protomaps-themes-base?activeTab=versions) to 1.3.1?
The last release of that package was 5 months ago.

PR #45, opened 2023-06-13
Add layer for country boundaries solid lines

cashc:
adds a `boundaries_country` layer for styling country boundaries with a solid line, as opposed to the dashed line that was used for all country/state/province boundaries

### DIFF
--- a/base/src/base_layers.ts
+++ b/base/src/base_layers.ts
@@ -798,10 +798,22 @@ export function nolabels_layers(source: string, c: Theme): any[] {
       },
     },
     {
+      id: "boundaries_country",
+      type: "line",
+      source: source,
+      "source-layer": "boundaries",
+      filter: ["<=", "pmap:min_admin_level", 2],
+      paint: {
+        "line-color": c.boundaries,
+        "line-width": 1,
+      },
+    },
+    {
       id: "boundaries",
       type: "line",
       source: source,
       "source-layer": "boundaries",
+      filter: [">", "pmap:min_admin_level", 2],
       paint: {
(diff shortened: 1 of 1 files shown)


nvkelso:
+1 for creating a separate color theme for these changes. Can that also be in a separate PR than the country lines change, please?

nvkelso:
Adding the original screenshots here for ease of reference:

before
image

after (ignore the thematic overlay and color changes)
image

nvkelso (approved):
LGTM

Thanks for splitting the color changes off to a separate PR :)

gzurbach:
@bdon is there anything we can do to move this forward? A check is failing but we're not sure if/how we should address it?

We would love to be able to use this new border styling in our maps :)

bdon:
@gzurbach thanks for being patient with this - right now I'm in a bit of a bind because:

I have merged this but you are still running "v2" of the tiles, correct? We're moving forward with a v3 of the tiles, but this means that our JS code to generate the style needs to match the breaking changes in tile v3. So having merged this into `main` it's difficult to get the changes back ported if you are including the style through the NPM package.

The most bulletproof way to work around this is to use JSON in your own application code to define the style, instead of relying on the JS package. Is that something you're doing now? 

The downside is that you'll miss out on our improvements to the styles in the near future, so we need to form a change-control strategy around this.

cashc:
@bdon yes, I believe we are still using "v2". And yes, we are using our own custom colors and not relying on the JS package currently. So I think our codebase in a good place. Thanks for the heads up on what migrating to v3 will involve!

PR #46, opened 2023-06-14
add BASIC theme for more standard map styling

cashc:
Adds a new theme `BASIC` that is a modified version of the LIGHT theme with colors to make roads, national forests, and bodies of water more obvious, among other small color tweaks that I think make the map more accessible and usable across a wide range of use cases.

See the map live at: https://beta.wwoofusa.org/en/hosts

### before
before

### after
after
contrast-zoomed


nvkelso:
Generally this looks good to me.

@bdon do you have opinions on the name of this color theme? Right now it's proposed as `BASIC`.

Do we also want to add a BASIC section to the final config object at bottom of the color file, too? Do these need to be registered anywhere else?

https://github.com/protomaps/basemaps/blob/f69d422c9dbc5827ae98e67dddff401c749a9219/base/src/colors.ts#L484-L490

eg, to:

```
 export default { 
     basic: BASIC, 
     light: LIGHT, 
     dark: DARK, 
     white: WHITE, 
     black: BLACK, 
     grayscale: GRAYSCALE, 
 }; 
```

base/index.html
```diff
@@ -15,6 +15,7 @@
             <div class="w-50 h-100 flex flex-column">
                 <div class="pa2">
                     <select v-model="selection">
+                        <option value="basic">Base Standard GL</option>
```
nvkelso, line 18: Hmm, `Base Basic GL` doesn't really roll off the tongue does it? But the `value` should match the name... so consider the awkward wording or coming up with a new name for your color schema (besides `BASIC`)?

base/src/colors.ts
```diff
@@ -67,6 +67,75 @@ export interface Theme {
     country_label_halo: string;
 }
 
+export const BASIC: Theme = {
```
nvkelso, line 70: Depending on naming decision, also update this const

base/src/colors.ts
```diff
@@ -413,6 +482,7 @@ export const GRAYSCALE: Theme = {
 };
 
 export default {
+    basic: BASIC,
```
nvkelso, line 485: Depending on naming decision, also update this export.

nvkelso (changes requested):
See comments, final name wrangling decisions...

cashc:
@nvkelso yeah, I agree the naming needs work. Could call it `STREET`?

base/src/colors.ts
```diff
@@ -67,6 +67,75 @@ export interface Theme {
     country_label_halo: string;
 }
 
+export const BASIC: Theme = {
+    hasCasings: true,
+    background: "#dddddd",
+    earth: "#ededed",
+    park: "#c2f7d1",
```
gzurbach, line 74: Consider using `#bfc99c` here. Otherwise parks stil use the bright green:

Screenshot 2023-06-16 at 22 52 10

cashc: done!

bdon:
`CONTRAST` is nice! 

To caveat this:

We're in the midst of reworking all the default styles, so I can't guarantee this additional style or any of the 5 existing styles is going to be preserved past the rework - but we DO want to make it easy to "theme" the map via a typed object like `Theme` right now. exact interface TBD. Thanks for PR!

cashc:
Understood! Thanks @bdon. Looking forward to seeing the upcoming changes and the TBD interface

PR #37, opened 2023-05-26
[WIP] Basic schema enhancements for Felt basemap

nvkelso:
@bdon this captures most of what we talked about Monday.

TODO:

- **Places layer:** `population_rank` isn't working as expected
- **Places layer:** we need a better value to use for client side sorting. Normally this would be the Natural Earth min_zoom (which solves problems like San Francisco having a smaller population than Oakland)
- Fussy linting nits

PR #47, opened 2023-06-16
Round 2 enhancements for Felt

nvkelso:
This is a followup to https://github.com/protomaps/basemaps/pull/37 which established basic changes needed for Felt's basemap.

**TL;DR for v3:**

- [x] This should get expanded into a CHANGELOG.md file at the root dir...

- Core Tilezen schema properties added:
    - `pmap:kind` is present on every feature now in every layer 
    - `pmap:kind_detail` is optionally present on some features in some layers
    - NOTE: in v4, the `pmap:` prefix will be removed (so `pmap:kind` will become simply `kind`)
- Core OSM tags for different kinds of places have been augmented, but...
    - `DEPRECATION WARNING`: These OSM tags are marked for deprecation in v4 schema, do not use these for styling. They aren't needed and take up extra file size.
    - If an explicate value is needed it should be a `kind`, or included in `kind_detail`. If there is a gap, please file an issue so it can be addressed.
- Zoom ranges of most features have been adjusted to remove details (and reduce file size) at early and mid-zooms. 
- Names have been removed from most features at early and mid-zooms (to also reduce file size)
    - Names have been kept on some features at early- and mid-zooms when they are known to be used for map display
- When features do have names, a `pmap:min_zoom` is added to achieve more predictable label collisions client side.
    - NOTE in v4, the `pmap:` prefix will be removed (so `pmap:min_zoom` will become simply `min_zoom`)
- Add Makefile with common build commands for easier development
- Improve consistency of internal private variable names in Planetiler profile.

**This PR will cover bug fixes and enhancements:**

- [x] **Places** layer bug fixes (`town`, `village` aren't `neighbourhood`; add `pmap:kind_detail` for original places tag values)
- [x] **Places** layer: should `population_rank` be prefixed with `pmap:`? Yes in v3.
- [x] **Places** layer: Fix the label grid sorting – Java strict type conversions are driving me nuts
    - [ ] Mid zooms are still loosing big cities like SF and San Jose, and Sacramento
- [ ] **Places** layer: Country label custom zooms ranges by named lists
    - [ ] this removes many country labels from early zooms when they couldn't reasonably be labeled anyhow (to reduce file size)
- [ ] **Places** layer: Region (eg state) label custom zooms ranges by named lists
    - [ ] this removes many region labels from early zooms when they couldn't reasonably be labeled anyhow (to reduce file size)
- [x] **POIs** layer: Better detection of `national_park` features (versus `park`) since you also have to look at `operator` tag to derive this from OSM 
    - [x] These should show up earlier / at all
- [x] **POIS** layer: derive label centroids from ways and relations features hugely increasing the number of included features compared to OSM source
- [x] **POIs** layer: add smattering of higher priority (even within a kind) features at earlier zoom levels (eg based on feature area)
    - [x] Remove the debug `pmap:area_debug` property
    - [x] Can the Seattle [space needle](https://www.openstreetmap.org/way/12903132) show up earlier? It's otherwise just an attraction shown at later zooms. Consider a variant of the `area` boost a `height` boost?
    - [ ] Push some zoom 14 parks later (node and small area, see SF)
- [x] **POIs** layer: Add pasthru unrestricted OSM values from `attraction`, `craft`, `historic`, `shop`, and `tourism` tags.
- [x] **Landuse/POIs** layers: Add allowlisted OSM values from natural (`beach`) and landuse (`cemetery`, `recreation_ground`, `winter_sports`, `quarry`, `park`, `forest`, `military`) tags.
- [x] **Landuse/Natural** layers: Better detection of `national_park` features (versus `park`) since you also have to look at `operator` tag to derive this from OSM
- [x] **Landuse/Natural** layers: Better detection of `forest` features (versus `wood`) since you also have to look at `operator` tag to derive this from OSM
- [x]  **Landuse/Natural** layers: Move `national_park`, `protected_area`, and `nature_reserve` to landuse layer as they are not natural but cultural. 
    - [x] This is breaking change bug fix from v2 to v3.
- [x] **Roads** layer: add `pmap:kind_detail` for values of `service` tag for `other` kind roads (eg for `parking_aisle` features)
- [x] **Roads** layer: change recently added `ref_length` to `shield_text_length` to more quickly converge towards Tilezen syntax
- [x] ~~**Roads** layer: highways with links are borked, investigate~~ Works fine in default style, debug Felt style.
- [x] **Roads** layer: Consider `brunnel` or `is_tunnel` / `is_bridge` booleans?
    - [x] `pmap:level` already does this?
- [x] ~~**Roads** layer: Consider turning off the link magic post processing (which would also keep the `kind_detail`) as it seems to be borking some highways?~~
- [x] **Roads** layer: show some roads at even earlier zooms (fix regression)
- [x] **Natural** layer: add same `pmap:kind` coallesce as in the Landuse layer
- [x] **Natural** layer: show some features at even earlier zooms (fix regression)
- [x] **Natural** layer: Add `grass` polygons.
- [x] **Natural** layer: show some features at even earlier zooms (fix regression)
- [x] **Natural Lines** layer: For linear water features like rivers – add tunnel / bridge indicator with `pmap:level` (same as roads layer) and intermittent indicators
- [x] **Natural points** layer: For label positions for water features like lakes
- [x] **Landuse** layer: show some features at even earlier zooms (fix regression)
- [x] **Landuse** layer: polygons should be merged (and not have names)
- [x]  **Landuse** layer: add explicate new `kind` values for: `beach`, `zoo`, `military`, `naval_base`, `airfield`
- [x]  **Landuse** layer:  Features with `leisure` tag are mapped to individual kind values instead of all erroneously to `park`.
- [x]  **Landuse** layer:  Features with `amenity` tag are mapped to individual kind values instead of all erroneously to `school` when not `hospital`.
- [x]  **Landuse** layer:  Features with `highway` IN `pedestrian` or `footway` are still exported as kind = `pedestrian` (as well `man_made` = `bridge` still mapping to kind = `pedestrian`)
- [x]  **Landuse** layer:  Add `pier` polygons.
- [x] **Physical points** layer: add derived lake (water) label position points
- [x] **Water** layer: better kinds to match new label points in physical points layer
    - [x] Reservoir should be `pmap:kind` = `water`, with the `kind_detail` set to `reservoir`?
- [x] **Water** layer: Use Natural Earth for low zoom water polygons
- [x] **Water** layer: Add tunnel / bridge indicator with `pmap:level` (same as roads layer) and intermittent, alkaline,  reservoir indicators
- [x] **Transit** layer: Add tunnel / bridge indicators with `pmap:level` (same as roads layer)
- [x] **Buildings** layer: Adjust zoom ranges, push building_park features to later zooms (to reduce file size)
- [x] **Buildings** layer: Quantize `height` tag at mid-zooms so more buildings merge in post-processing (to reduce file size)
- [x] **Buildings** layer: Add optional `min_height` tag (for 2.5D and 3D visualizations)
- [ ] **Boundaries** layer: QA/QC fixes for the initial support for "disputed" and related tags
- [ ] **Boundaries** layer: QA/QC the existing approach for min admin_level ... in some situations might need to always export say `locality` (city) boundary lines (even if coterminous with parent placetypes)
- [x] **Boundaries** layer: Don't export admin_level 1, 3, 5, and 7 – those generally aren't styled and are taking up a lot of file size even at very low zooms
- [x] Update default Protomaps style so it shows off the new features (and any impactful changes)

Followup in separate PR?

- [ ] Add localization configuration for how many name locales are exported in tiles (this is a large driver of tile file size at low and mid zooms)


nvkelso:
Will fix https://github.com/protomaps/basemaps/issues/48.

tiles/src/main/java/com/protomaps/basemap/Basemap.java
```diff
@@ -50,6 +50,7 @@ public Basemap(Boolean useTilezen) {
       var physical_point = new PhysicalPoint();
       registerHandler(physical_point);
       registerSourceHandler("osm", physical_point);
+      registerSourceHandler("ne", physical_point::processNe);
```
nvkelso, line 53: To enable lake label generation from Natural Earth at low zooms

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -38,10 +40,27 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
           line.setMinZoom(0);
         } else if (minAdminLevel.getAsInt() <= 4) {
           line.setMinZoom(3);
+        } else if (minAdminLevel.getAsInt() <= 6) {
```
nvkelso, line 43: Show county lines at earlier zoom

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -38,10 +40,27 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
         } else {
           line.setMinZoom(10);
         }
 
+        switch (minAdminLevel.getAsInt()) {
+          case 2 -> {
+            line.setAttr("pmap:kind", "country");
+          }
+          case 4 -> {
+            line.setAttr("pmap:kind", "region");
```
nvkelso, line 54: This sets Tilezen kind values – so `region` instead of `state`/`province` and `locality` instead of `city`.

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -20,10 +21,33 @@ public String name() {
...
+    }
+
+    return (int) Math.round(val / step) * step;
+  }
+
   @Override
   public void processFeature(SourceFeature sf, FeatureCollector features) {
-    if (sf.canBePolygon() && (sf.hasTag("building") || sf.hasTag("building:part"))) {
+    if (sf.canBePolygon() && (
+            ( sf.hasTag("building") && !sf.hasTag("building", "no")) ||
```
nvkelso, line 37: Sometimes `building` (and `building:part`) are marked `no`, we should exclude those from tiles.

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -20,10 +21,33 @@ public String name() {
...
+            ( sf.hasTag("building:part") && !sf.hasTag("building:part", "no"))))
+    {
       Double height = parseDoubleOrNull(sf.getString("height"));
+      Double min_height = parseDoubleOrNull(sf.getString("min_height"));
```
nvkelso, line 41: Building parts start offset from the ground level, for 3D maps

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -20,10 +21,33 @@ public String name() {
...
+      Integer min_zoom = 11;
+      String kind = "building";
+
+      // Limit building:part features to later zooms
+      // TODO: (nvkelso 20230621) this should be based on area and volume, too
+      if( sf.hasTag("building:part") ) {
+        kind = "building_part";
+        min_zoom = 14;
```
nvkelso, line 49: Building parts aren't useful to show until alter zooms when they become large enough on the map to discern.

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -34,12 +58,21 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
+      if( kind == "building_part") {
+        // We don't need to set WithMinzoom because that's implicate with the ZoomRange
+        feature.setAttr("pmap:kind_detail", sf.getString("building:part"));
+        feature.setAttr("min_height", sf.getString("min_height"));
+      }
 
-      OsmNames.setOsmNames(feature, sf, 13);
+      // Names should mostly just be for POIs
+      // Sometimes building name and address are useful items, but only at zoom 17+
+      //OsmNames.setOsmNames(feature, sf, 13);
```
nvkelso, line 75: Drop the names from buildings... if it's really important it'll show up in the POIs layer.

Alternatively we could only set this starting at zoom 15? As some building names will be lost (when there isn't a business there).

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -50,8 +83,62 @@ public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> i
...
     items = Area.filterArea(items, 0);
 
-    if (zoom >= 14)
+    // DEBUG
+    //items = Area.addAreaTag(items);
+
+    if (zoom >= 15)
       return items;
+
+    // quantize height by zoom when less than max_zoom 15 to facilitate better feature merging
```
nvkelso, line 92: Buildings weren't really merging much before because of all the unique values. Quantizing them encourages more merges.

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
```diff
@@ -17,17 +17,18 @@ public String name() {
 
   public void processOsm(SourceFeature sf, FeatureCollector features) {
     features.polygon(this.name())
-      .setZoomRange(6, 15).setBufferPixels(8);
+        .setAttr("pmap:kind", "earth")
```
nvkelso, line 20: This is superfluous but nice to have in prep for v4 spec so each feature has a kind. They get merged, so very small file size hit.

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
     if (sf.canBePolygon() && (sf.hasTag("aeroway", "aerodrome", "runway") ||
       sf.hasTag("area:aeroway", "taxiway", "runway") ||
       sf.hasTag("amenity", "hospital", "school", "kindergarten", "university", "college") ||
+      sf.hasTag("boundary", "national_park", "protected_area") ||
```
nvkelso, line 22: Breaking change in v2 > v3: `national_park` and `protected_area` moved to landuse layer form natural layer.

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
       sf.hasTag("landuse", "recreation_ground", "industrial", "brownfield", "railway", "cemetery", "commercial",
         "grass", "orchard", "farmland", "farmyard", "residential") ||
-      sf.hasTag("leisure", "park", "garden", "golf_course", "dog_park", "playground", "pitch") ||
+      sf.hasTag("leisure", "park", "garden", "golf_course", "dog_park", "playground", "pitch", "nature_reserve") ||
```
nvkelso, line 25: Breaking change in v2 > v3: `nature_reserve` moved to landuse layer form natural layer.

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
       sf.hasTag("man_made", "pier") ||
+      // TODO: (nvkelso 20230622) This use of the place tag here is dubious
       sf.hasTag("place", "neighbourhood") ||
```
nvkelso, line 28: @bdon do you know the history of this tag usage here? It seems like a bug...

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
+        (sf.hasTag("highway", "pedestrian", "footway") || sf.hasTag("man_made", "bridge")))))
+    {
       var poly = features.polygon(layerName)
         .setId(FeatureId.create(sf))
-        .setAttr("landuse", sf.getString("landuse"))
-        .setAttr("leisure", sf.getString("leisure"))
         .setAttr("aeroway", sf.getString("aeroway"))
-        .setAttr("area:aeroway", sf.getString("area:aeroway"))
         .setAttr("amenity", sf.getString("amenity"))
+        .setAttr("area:aeroway", sf.getString("area:aeroway"))
```
nvkelso, line 37: a-Z sorting for cleaner future diffs

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
 
-      OsmNames.setOsmNames(poly, sf, 0);
-
-      if (ghostFeatures) {
-        poly.setAttr("isGhostFeature", true);
-      }
-
-      poly.setAttr("pmap:area", "");
+      // NOTE: (nvkelso 20230622) landuse labels for polygons are found in the pois layer
+      //OsmNames.setOsmNames(poly, sf, 0);
```
nvkelso, line 52: No longer export names on the landuse polygons, instead we peal off named features for the POIs layer.

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
-        }
-      } else if (sf.hasTag("landuse")) {
-        if (sf.hasTag("landuse", "orchard", "farmland", "farmyard")) {
-          kind = "farmland";
-        } else if (sf.hasTag("landuse", "industrial", "brownfield")) {
-          kind = "industrial";
-        } else if (sf.hasTag("landuse", "cemetery")) {
-          kind = "cemetery";
-        }
+      if (sf.hasTag("aeroway", "aerodrome")) {
```
nvkelso, line 55: Major refactor of explicate and implicate kind values

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -18,61 +19,105 @@ public static void processFeature(SourceFeature sf, FeatureCollector features, S
...
+        } else if (sf.hasTag("tourism")) {
+          kind = sf.getString("tourism");
+        // Boundary is most generic, so place last else we loose out
+        // on nature_reserve detail versus all the protected_area
+        } else if (sf.hasTag("boundary")) {
+          kind = sf.getString("boundary");
+        }
       }
+
+      if( sf.hasTag("boundary", "national_park") &&
```
nvkelso, line 107: Special case `national_park` kind calcuation.

We should do this for `national_forest`, too...

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -89,7 +134,7 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
 
   @Override
   public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> items) throws GeometryException {
-    items = Area.addAreaTag(items);
+    //items = Area.addAreaTag(items);
```
nvkelso, line 137: ¿Do I need this for the later filter to work?

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -98,6 +143,7 @@ public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> i
     else if (zoom <= 5)
       minArea = 800 / (4096 * 4096) * (256 * 256);
     items = Area.filterArea(items, minArea);
-    return items;
+
+    return FeatureMerge.mergeNearbyPolygons(items, 3.125, 3.125, 0.5, 0.5);
```
nvkelso, line 147: Merge early and mid-zoom polygons for landuse layer

tiles/src/main/java/com/protomaps/basemap/layers/Natural.java
```diff
@@ -20,32 +21,44 @@ public String name() {
   @Override
   public void processFeature(SourceFeature sf, FeatureCollector features) {
     if (sf.canBePolygon() && (sf.hasTag("natural", "wood", "glacier", "scrub", "sand", "wetland", "bare_rock") ||
-      sf.hasTag("landuse", "forest", "meadow") || sf.hasTag("leisure", "nature_reserve") ||
-      sf.hasTag("boundary", "national_park", "protected_area"))) {
```
nvkelso, line 24: Moved to landuse layer

tiles/src/main/java/com/protomaps/basemap/layers/Natural.java
```diff
@@ -20,32 +21,44 @@ public String name() {
...
+
+      feat.setAttr("pmap:kind", kind);
+
     }
   }
 
   @Override
   public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> items) throws GeometryException {
-    items = Area.addAreaTag(items);
+    //items = Area.addAreaTag(items);
```
nvkelso, line 51: Do I need this for later filterArea function?

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalLine.java
```diff
@@ -18,30 +18,36 @@ public String name() {
...
     if (sf.canBeLine() && (sf.hasTag("waterway") ||
-      sf.hasTag("natural", "strait", "cliff")) && (!sf.hasTag("waterway", "riverbank", "reservoir"))) {
+      sf.hasTag("natural", "strait", "cliff")) && (!sf.hasTag("waterway", "riverbank", "reservoir")))
+    {
+      int min_zoom = 12;
+      String kind = "other";
+      if (sf.hasTag("waterway")) {
+        kind = sf.getString("waterway");
+        if (sf.hasTag("waterway","river")) {
+          min_zoom = 9;
```
nvkelso, line 28: Show river lines earlier (than stream)

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalLine.java
```diff
@@ -18,30 +18,36 @@ public String name() {
...
+        }
+      } else if (sf.hasTag("natural")) {
+        kind = sf.getString("natural");
+      }
+
       var feat = features.line(this.name())
         .setId(FeatureId.create(sf))
+        // Core Tilezen schema properties
+        .setAttr("pmap:kind", kind)
```
nvkelso, line 37: Move towards Tilezen convention of single coalesce property across all OSM tags.

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalLine.java
```diff
@@ -18,30 +18,36 @@ public String name() {
...
         // Add less common attributes only at higher zooms
         .setAttrWithMinzoom("bridge", sf.getString("bridge"), 12)
         .setAttrWithMinzoom("tunnel", sf.getString("tunnel"), 12)
         .setAttrWithMinzoom("layer", sf.getString("layer"), 12)
-        .setZoomRange(12, 15);
+        .setZoomRange(min_zoom, 15);
 
       if (sf.hasTag("intermittent", "yes")) {
-        feat.setAttr("intermittent", 1);
+        feat.setAttr("intermittent", true);
```
nvkelso, line 48: ¿Decide if boolean or `1` value?

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalPoint.java
```diff
@@ -15,33 +20,213 @@ public String name() {
...
       // TODO: rank based on ele
 
+      String kind = "";
+
       int minzoom = 12;
-      if (sf.hasTag("natural", "peak")) {
-        minzoom = 13;
+      if (sf.hasTag("place", "ocean")) {
+        kind = "ocean";
+        minzoom = 0;
```
nvkelso, line 74: Show `ocean` earlier than `sea`.

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalPoint.java
```diff
@@ -15,33 +20,213 @@ public String name() {
...
         minzoom = 3;
       }
+      if (sf.hasTag("natural", "peak")) {
+        kind = "peak";
+        minzoom = 13;
+      }
 
       var feat = features.point(this.name())
         .setId(FeatureId.create(sf))
+        .setAttr("pmap:kind", kind)
```
nvkelso, line 87: Move towards Tilezen standard of always exporting coallesed `kind` value from OSM tags

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalPoint.java
```diff
@@ -15,33 +20,213 @@ public String name() {
...
         .setAttr("place", sf.getString("place"))
         .setAttr("natural", sf.getString("natural"))
         .setAttr("ele", sf.getString("ele"))
         .setZoomRange(minzoom, 15);
 
       OsmNames.setOsmNames(feat, sf, 0);
     }
+
+    if( sf.hasTag("name") && sf.getTag( "name") != null &&
```
nvkelso, line 96: Start of water polygon label centroid logic...

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalPoint.java
```diff
@@ -15,33 +20,213 @@ public String name() {
...
+      } catch(GeometryException e) {
+        System.out.println(e);
+      }
+
+      var name_min_zoom = 15;
+      var kind_detail = "";
+      var reservoir = false;
+      var alkaline = false;
+
+      // coallese values across tags to single kind value
```
nvkelso, line 120: New logic for calculating `kind` which more closely matches Tilezen v1.9

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalPoint.java
```diff
@@ -15,33 +20,213 @@ public String name() {
...
+        kind = sf.getString("waterway");
+      } else if (sf.hasTag("landuse", "basin", "reservoir")) {
+        kind = sf.getString("landuse");
+      } else if (sf.hasTag("leisure", "swimming_pool")) {
+        kind = "swimming_pool";
+      } else if (sf.hasTag("amenity", "swimming_pool")) {
+        kind = "swimming_pool";
+      }
+
+      // We don't want to show too many water labels at early zooms else it crowds the map
```
nvkelso, line 152: Show larger lake's labels earlier than smaller lake's labels...

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -19,6 +23,22 @@ public String name() {
     return "places";
   }
 
+  // Evaluates place layer sort ordering of inputs into an integer for the sort-key field.
+  /*
+  static int getSortKey(float min_zoom, int population_rank, long population, String name) {
```
nvkelso, line 28: @bdon Newbie Java guy here, super confused by all the type conversions and related errors, so this is pseudocode "working" but various compiler errors about type conversions...

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -37,14 +57,14 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
         case "Admin-0 capital":
         case "Admin-0 capital alt":
         case "Admin-0 region capital":
-          kind = "city";
+          kind = "locality";
```
nvkelso, line 60: WARNING: Breaking v2 > v3 Protomaps schema change to converge on Tilezen kind values for `locality` (instead of `city`, which is still in `kind_detail`).

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -66,10 +86,13 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
           theme_max_zoom)
         .setAttr("pmap:kind", kind)
         .setAttr("pmap:kind_detail", kind_detail)
-        .setAttr("population", sf.getString("pop_max"))
-        .setAttr("population_rank", sf.getString("rank_max"))
+        .setAttr("population", parseIntOrNull(sf.getString("pop_max")))
+        .setAttr("pmap:population_rank", parseIntOrNull(sf.getString("rank_max")))
```
nvkelso, line 90: This is a "new" property for Protomaps schema v3 spec, but it's not OSMy so let's prefix it with `pmap:` here for consistency with other pmap specific properies... and in v4 we'd remove the prefix.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -66,10 +86,13 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
...
         .setAttr("wikidata_id", sf.getString("wikidata"))
-        .setBufferPixels(128);
+        .setBufferPixels(128)
+        // we set the sort keys so the label grid can be sorted predictably (bonus: tile features also sorted)
+        //.setSortKey(getSortKey("pmap:min_zoom",  "population_rank", "population", "name"))
+        .setPointLabelGridPixelSize(12, 128);
```
nvkelso, line 95: We are using a label grid now (good)

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -66,10 +86,13 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
...
         .setAttr("pmap:kind_detail", kind_detail)
-        .setAttr("population", sf.getString("pop_max"))
-        .setAttr("population_rank", sf.getString("rank_max"))
+        .setAttr("population", parseIntOrNull(sf.getString("pop_max")))
+        .setAttr("pmap:population_rank", parseIntOrNull(sf.getString("rank_max")))
         .setAttr("wikidata_id", sf.getString("wikidata"))
-        .setBufferPixels(128);
+        .setBufferPixels(128)
+        // we set the sort keys so the label grid can be sorted predictably (bonus: tile features also sorted)
+        //.setSortKey(getSortKey("pmap:min_zoom",  "population_rank", "population", "name"))
```
nvkelso, line 94: @bdon once the sort key function and Java type conversions are fixed, this can be re-enabled.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -89,45 +113,50 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
-      if (sf.hasTag("place", "country")) {
+      if (place.equals("country")) {
         feat.setAttr("pmap:kind", "country")
+          // TODO: these should be from data join to Natural Earth, and if fail data join then default to 8
+          .setAttr("pmap:min_zoom", 1)
           .setZoomRange(0, 9);
-      } else if (sf.hasTag("place", "state", "province")) {
-        feat.setAttr("pmap:kind", "state")
+      } else if (place.equals("state") || place.equals("province")) {
+        feat.setAttr("pmap:kind", "region")
```
nvkelso, line 122: Similarly Tilezen forward, `region` instead of `state`.

And `locality` instead of `city` below...

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -158,7 +187,17 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
 
-      feat.setAttr("population_rank", population_rank);
+      feat.setAttr("pmap:population_rank", population_rank);
+
+      if (population > 0) {
+        feat.setAttr("population", population);
+
+        // we set the sort keys so the label grid can be sorted predictably (bonus: tile features also sorted)
+        feat.setPointLabelGridPixelSize(12, 128);
+        //feat.setSortKey(getSortKey("pmap:min_zoom",  "pmap:population_rank", "population", "name"));
```
nvkelso, line 197: @bdon 2nd place to re-enable the sort key stuff once Java type conversion problems are fixed

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
+
   @Override
   public void processFeature(SourceFeature sf, FeatureCollector features) {
-    if (!sf.canBeLine() && (sf.isPoint() || sf.canBePolygon()) && (sf.hasTag("amenity") ||
-      sf.hasTag("shop") ||
-      sf.hasTag("tourism") ||
-      sf.hasTag("leisure") ||
-      sf.hasTag("aeroway", "aerodrome") ||
-      sf.hasTag("railway", "station"))) {
+    if ( (sf.isPoint() || sf.canBePolygon()) && (
```
nvkelso, line 33: Significant reworking of what's eligible to be a POI...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
+            sf.hasTag("leisure") ||
+            sf.hasTag("railway", "station") ||
+            sf.hasTag("shop") ||
+            sf.hasTag("tourism") ))
+    {
       String kind = "other";
+      String kind_detail = "";
+      Integer min_zoom = 15;
+
       if (sf.hasTag("aeroway", "aerodrome")) {
```
nvkelso, line 49: Significant reworking of `kind` calculations...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
-        // This is dubious but existing behavior
-        kind = "park";
+        min_zoom = 13;
       } else if (sf.hasTag("shop", "grocery", "supermarket")) {
         kind = sf.getString("shop");
+        min_zoom = 14;
       } else if (sf.hasTag("tourism", "attraction", "camp_site", "hotel")) {
         kind = sf.getString("tourism");
+        min_zoom = 15;
+      } else {
```
nvkelso, line 85: If it's not in the allowlist, still let it thru but with a deferred min zoom of 15.

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
+        } else if (sf.hasTag("tourism")) {
+          kind = sf.getString("tourism");
+        // Boundary is most generic, so place last else we loose out
+        // on nature_reserve detail versus all the protected_area
+        } else if (sf.hasTag("boundary")) {
+          kind = sf.getString("boundary");
+        }
+      }
+
+      if( sf.hasTag("boundary", "national_park") &&
```
nvkelso, line 115: Special logic for `national_park`.

We'd also want handling for `national_forest`...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
+        } catch(GeometryException e) {
+          System.out.println(e);
+        }
+
+        Double height = 0.0;
+        try { height = sf.getString("height") == null ? 0.0 : parseDoubleOrNull(sf.getString("height")); } catch(Exception e) {
+          System.out.println("Problem getting height");
+        }
+
+        // Area zoom grading overrides the kind zoom grading in the section above.
```
nvkelso, line 152: If the building or landuse polygon was large, then show the POI earlier than default for nodes...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
+        } else {
+          if (way_area >   10  ) {            //    500000
+            min_zoom = 11;
+          } else if (way_area >    2  ) {     //     50000
+            min_zoom = 12;
+          } else if (way_area >    0.5) {     //     10000
+            min_zoom = 13;
+          }
+
+          // Small but tall features should show up early as they have regional prominance.
```
nvkelso, line 199: If the building is super tall, then show the POI earlier...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -3,85 +3,308 @@
...
 
+        // Core Tilezen schema properties
+        if( kind_detail != "") {
+          point_feature.setAttr("pmap:kind_detail", kind_detail);
+        }
+
         OsmNames.setOsmNames(point_feature, sf, 0);
 
-        point_feature.setAttr("pmap:kind", kind);
+        // Some features should only be visible at very late zooms when they don't have a name
```
nvkelso, line 283: There were still too many features at zoom 15, so annotate some for later zooms. 

Should be combined with a style layer filter and style collision property.

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -35,7 +35,7 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
           shield_text = shield_text.replaceAll("I ", "");
           network_val = "US:I";
         } else {
-          network_val = "other";
+          network_val = network_val;
```
nvkelso, line 38: This will allow other shields besides `US:I` and `US:US` instead of blocking them...

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -52,15 +52,17 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
       if (highway.equals("motorway") || highway.equals("motorway_link")) {
-        feat.setAttr("pmap:kind", "highway").setZoomRange(6, 15);
+        // TODO: (nvkelso 20230622) Use Natural Earth for low zoom roads at zoom 5 and earlier
+        //       as normally OSM roads would start at 6, but we start at 3 to match Protomaps v2
+        feat.setAttr("pmap:kind", "highway").setZoomRange(3, 15);
 
         if (highway.equals("motorway")) {
           feat.setAttrWithMinzoom("ref", shield_text, 7)
-            .setAttrWithMinzoom("ref_length", shield_text_length, 7)
+            .setAttrWithMinzoom("shield_text_length", shield_text_length, 7)
```
nvkelso, line 61: Switch to Tilezen forward naming on the new v3 property for `shield_text_length` (instead of `ref_length`).

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -94,38 +98,64 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
         OsmNames.setOsmNames(feat, sourceFeature, 13);
       } else if (highway.equals("residential") || highway.equals("service") || highway.equals("unclassified") ||
         highway.equals("road") || highway.equals("raceway")) {
         feat.setAttr("pmap:kind", "minor_road").setZoomRange(12, 15)
           .setAttrWithMinzoom("ref", shield_text, 12)
-          .setAttrWithMinzoom("ref_length", shield_text_length, 12)
+          .setAttrWithMinzoom("shield_text_length", shield_text_length, 12)
           .setAttrWithMinzoom("network", network_val, 12);
+
+        if( highway.equals("service") ) {
```
nvkelso, line 120: Defer showing small and busy service roads until later, and later later.

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -94,38 +98,64 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
+        feat.setAttr("pmap:kind", "path").setZoomRange(12, 15)
+          .setAttr("pmap:kind_detail", highway).setZoomRange(12, 15)
           .setAttrWithMinzoom("ref", shield_text, 12)
-          .setAttrWithMinzoom("ref_length", shield_text_length, 12)
-          .setAttrWithMinzoom("network", network_val, 12)
-          .setAttr("pmap:kind", "path").setZoomRange(12, 15);
+          .setAttrWithMinzoom("shield_text_length", shield_text_length, 12)
+          .setAttrWithMinzoom("network", network_val, 12);
+
+        if( sourceFeature.hasTag("highway", "path", "cycleway", "bridleway", "footway", "steps") ) {
```
nvkelso, line 140: Defer showing small and busy paths until later, and later later.

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -146,12 +176,14 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
 
   @Override
   public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> items) {
+    if( zoom < 12 ) {
```
nvkelso, line 179: The linkSimplify logic is still mysterious to me... either it has a bug or it shouldn't be exercised at higher zooms (or both!?), else links were totally missing at mid-zooms

tiles/src/main/java/com/protomaps/basemap/layers/Transit.java
```diff
@@ -27,38 +27,74 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
 
       int minzoom = 11;
 
+      if (sf.hasTag("aeroway", "runway")) {
```
nvkelso, line 30: Airports are important, show these runway and taxiway lines earlier

tiles/src/main/java/com/protomaps/basemap/layers/Transit.java
```diff
@@ -27,38 +27,74 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
+        minzoom = 9;
+      } else
+      if (sf.hasTag("aeroway", "taxiway")) {
+        minzoom = 10;
+      } else
       if (sf.hasTag("service", "yard", "siding", "crossover")) {
         minzoom = 13;
+      } else
+      if (sf.hasTag("man_made", "pier")) {
```
nvkelso, line 39: Early piers were cluttering up the map, so push them back...

tiles/src/main/java/com/protomaps/basemap/layers/Transit.java
```diff
@@ -27,38 +27,74 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
-        .setAttr("service", sf.getString("service"))
-        .setAttr("aerialway", sf.getString("aerialway"))
-        .setAttr("network", sf.getString("network"))
-        .setAttr("ref", sf.getString("ref"))
-        .setAttr("highspeed", sf.getString("highspeed"))
-        .setAttr("layer", sf.getString("layer"))
-        .setZoomRange(minzoom, 15);
-
       String kind = "other";
+      String kind_detail = "other";
```
nvkelso, line 44: More sophisticated `kind` calculation...

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
```diff
@@ -51,25 +102,83 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
         .setMinPixelSize(3.0)
         .setBufferPixels(8);
 
+      String kind = "other";
+      String kind_detail = "";
+      var reservoir = false;
+      var alkaline = false;
+
+      // coallese values across tags to single kind value
+      if (sf.hasTag("natural", "water", "bay", "strait", "fjord")) {
```
nvkelso, line 111: Use the same upgraded `kind` logic across water, physical lines, and physical points (label centroids)

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
```diff
@@ -51,25 +102,83 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
+      }
+
+      feature.setAttr("pmap:kind", kind);
+
+      // Optional tags
+      if (kind_detail != "") {
+        feature.setAttr("pmap:kind_detail", kind_detail);
+      }
+      if (sf.hasTag("water", "reservoir") || reservoir) {
+        feature.setAttr("reservoir", true);
```
nvkelso, line 154: @bdon Do you want these to be booleans or `1` int values?

nvkelso:
@bdon This is ready for initial review, thanks!

In the **Places** layer I could use your help on setting up the feature sorting. As a Java newbie I got super confused with Java strict type conversions, maybe you'll have better luck?

The disputed boundaries stuff I'll file a separate PR for if it's actually needed...

tiles/Makefile
```diff
@@ -0,0 +1,61 @@
+# download dependencies and compile the JAR
```
nvkelso, line 1: New Makefile with easier to type commands...

tiles/Makefile
```diff
@@ -0,0 +1,61 @@
...
+	rm -rf *.pmtiles
+
+# This is optimized for local dev serving PMTiles out of directory local to this Makefile
+# Don't use this for production (instead set --cors=ORIGIN)
+# The default port is: 8080 (use --port to override)
+# Assumes go-pmtiles has been installed locally (eg prebuilt go binary download) & the path is added to your shell env
+# https://github.com/protomaps/go-pmtiles
+# Example tile coord to test with: http://localhost:8080/monaco/12/2133/1495.mvt
+serve:
+	~/Downloads/pmtiles serve --cors=* .
```
nvkelso, line 17: TODO: This path should be updated...

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -35,98 +43,134 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
+        min_zoom_names = 14;
+      }
+
+      var feat = features.line("roads")
+                  .setId(FeatureId.create(sourceFeature))
+                  // Core Tilezen schema properties
+                  .setAttr("pmap:kind", kind)
+                  .setAttr("pmap:kind_detail", kind_detail)
+                  .setAttrWithMinzoom("ref", shield_text, min_zoom_shield_text)
+                  .setAttrWithMinzoom("shield_text_length", shield_text_length, min_zoom_shield_text)
```
nvkelso, line 152: Switch to Tilezen forward naming on the new v3 property for shield_text_length (instead of ref_length).

_NOTE: This isn't a breaking change because we haven't tagged v3 yet (even though `ref_length` was merged for a few weeks._

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
nvkelso: It produces gaps at mid-zooms still

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -35,98 +43,134 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
-          .setAttrWithMinzoom("network", network_val, 12);
-        OsmNames.setOsmNames(feat, sourceFeature, 14);
+        highway.equals("road") || highway.equals("raceway")
+      ) {
+        kind = "minor_road";
+        min_zoom = 12;
+        min_zoom_shield_text = 12;
+        min_zoom_names = 14;
+
+        if (highway.equals("service")) {
```
nvkelso, line 109: Defer showing small and busy service roads until later, and later later.

compare/Makefile
```diff
@@ -0,0 +1,4 @@
+serve:
```
nvkelso, line 1: This is already in the README so this is a convenience.

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -21,29 +21,213 @@ public String name() {
...
+    if (sf.canBeLine() && sf.hasTag("min_zoom") && (kind.equals("") == false && kind.equals("tz_boundary") == false)) {
+      features.line(this.name())
+        // Don't label lines to reduce file size (and they aren't shown in styles anyhow)
+        //.setAttr("name", sf.getString("name"))
+        .setAttr("pmap:min_zoom", sf.getLong("min_zoom"))
+        .setAttr("pmap:min_admin_level", admin_level)
+        .setZoomRange(
+          sf.getString("min_zoom") == null ? theme_min_zoom : (int) Double.parseDouble(sf.getString("min_zoom")),
+          theme_max_zoom)
+        .setAttr("pmap:ne_id", sf.getString("ne_id"))
```
bdon, line 157: I would prefer to embed ne or OSM IDs in the 64-bit MVT id - are we ever going to use this for visualization or joining?

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -21,29 +21,213 @@ public String name() {
...
+        }
+
+        if (kind != "" && kind_detail != "") {
+          var line = features.line(this.name())
+            .setId(FeatureId.create(sf))
+            .setMinPixelSize(0)
+            .setAttr("pmap:min_admin_level", minAdminLevel.getAsInt())
+            .setAttr("pmap:kind", kind)
+            .setAttr("pmap:kind_detail", kind_detail)
+            .setAttr("pmap:min_zoom", min_zoom)
```
bdon, line 224: So as is we will have a feature with `pmap:min_zoom=0` that starts appearing at 6, is this to enable tag continuity with the NE feature from zooms 0-5? Or used as a proxy for label weight/size?

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
bdon: You don't  - the reason it exists was at some point style-side filtering was happening based on area. If Tilezen doesn't need something like this, we can remove it.

tiles/src/main/java/com/protomaps/basemap/layers/Natural.java
bdon: Nope! Just slaps a tag on each feature, the filterArea step recomputes it.

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,200 @@
...
+#### Developer level of effort:
+
+- Major version X: **high** – significant integration challenges, read the changelog closely
+- Minor version Y: **low** – some integration challenges, read the changelog
+- Bug fixes Z: **none** – simply use the new tiles, skim or ignore the changelog
+- Pre-release POSTFIX: **low to high** – some to significant integration challenges, read the changelog
+
+## Versioning the Protomaps Basemap vector tile schema
+
+Upon our version `1.0.0` release Protomaps Basemap makes the following promises based on the Tilezen schema:
```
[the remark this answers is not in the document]
nvkelso: This is largely a copy-paste job from Tilezen, so there may be some details (some of which you've noted) to iron out still. Thanks for being thorough here...

SemVer is pretty explicate that once you hit 1.0.0 you've made a contract with the end-user with respect to future major-minor-patch versions.

So even though I'm conceptualizing this PR as being for Protomaps version `3.0.0` schema, I think this doc should still talk about `1.0.0`.

Your description of how Protomaps API works is _related_ to the version of the `schema`. Your description of `/v2/` and `/v2/20230701/` make sense to me. 

I don't have strong opinions about how you setup the API, though I think it's useful to think about it in SemVer terms:

- Next major version: `/v3/20230701-beta/` or `/v3/beta-v3.0.0-pre-20230701/`
- Next minor version: `/v2/20230701-beta/` or `/v2/beta-v2.5.0-pre-20230701/`
- Next patch version: `/v2/20230701-beta/` or `/v2/beta-v2.5.1-pre-20230701/`

Where `20230701-beta` is an arbitrary build date (and allow for an {a,b,c} variant) with text indicator that it's not production worthy. So the question is where to put metadata about what schema (pre)release or SHA for the code was used, what OSM data went in, what Natural Earth version was used – and in future what Who's On First version (and even a landcover source)... 

All the beta releases shouldn't have any user facing contract – the are effectively ephemeral. But both incremental minor released versions and patch released versions and beta versions need to be allowed for under the same "`/v2/`" path and depending on developer velocity sometimes they may all be in flight at the same time.

l'll note here other discussion elsewhere in this repo about NPM package is more about the `style`, which is somewhat dependent on the schema. Which creates an awkward situation of having 2 different "versions" in the same repo. Hmm.

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+### Frequency of data updates
+
+Tilezen has 4 primary sources:
+
+- **Natural Earth** (used for zooms 0-8 for most everything) updates infrequently (often annually)
+- **OpenStreetMap** (used for zooms 9+ for most everything, sometimes earlier) updates frequently (at least daily)
+- **OpenStreetMapData** (used for zooms 9+ in the earth and water layers only) updates infrequently (optimistically monthly)
+- **Who’s On First** (used for zooms 12+ for places layer) updates frequently (at least daily)
```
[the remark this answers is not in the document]
nvkelso: WOF is not yet incorporated here in v3, that's a v4 task (or v3.1 if really ambitious). So this line should be removed or rephrased now.

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+### INDIVIDUAL FEATURES are not versioned
```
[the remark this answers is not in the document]
nvkelso: +1 for updating the metadata to be more exhaustive.

tiles/Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+#
+# Testing areas
+#
+
+# Smallest
+#
+# Download and generate monaco.pmtiles in the current directory:
+monaco:
```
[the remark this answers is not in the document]
nvkelso: Generally I look for a balance of urban, suburban, and rural across a range of different regions around the world to capture different IRL density of people and built structures, and the "fill rate" of OSM to match that IRL ideal, and with different character sets for localization.

Off the cuff, something like?

- California (whole state or really wide square cut of SF to Tahoe)
- New York, NY (metro area)
- Switzerland (whole country)
- London (metro area)
- Paris (metro area)
- Berlin (metro area)
- Tokyo (metro area)
- Mumbai (metro area)
- Riyad (metro area)
- Jakarta (metro area)
- Cairo (metro area)
- Rio de Janeiro (metro area)

tiles/src/main/java/com/protomaps/basemap/feature/CountryNameZooms.java
```diff
@@ -0,0 +1,1216 @@
+package com.protomaps.basemap.feature;
+
+import com.onthegomap.planetiler.reader.SourceFeature;
+
+public class CountryNameZooms {
+
+    public static float[] getMinMaxZooms(SourceFeature sf) {
```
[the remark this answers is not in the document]
nvkelso: Sure! I'm still a newbie with Java so a more performant data structure would be great now I've proven the logic works.

tiles/src/main/java/com/protomaps/basemap/feature/CountryNameZooms.java
```diff
@@ -0,0 +1,1216 @@
...
+        float min_zoom = 8.0f;   // default for unrecognized countries
+        float max_zoom = 11.0f;  // default for all countries
+
+        try {
+            var name = sf.getString("name:en") == null ? sf.getString("name") : sf.getString("name:en");
+
+            if (name != null) {
+                switch (name) {
+                    case "Andorra" -> {
+                        // AD country code
```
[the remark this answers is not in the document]
nvkelso: In the past ISO codes weren't populated reliably, but that seems to be shifting in OSM. I'm open to doing either, as long there's good "recall".

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -21,29 +21,213 @@ public String name() {
...
+      features.line(this.name())
+        // Don't label lines to reduce file size (and they aren't shown in styles anyhow)
+        //.setAttr("name", sf.getString("name"))
+        .setAttr("pmap:min_zoom", sf.getLong("min_zoom"))
+        .setAttr("pmap:min_admin_level", admin_level)
+        .setZoomRange(
+          sf.getString("min_zoom") == null ? theme_min_zoom : (int) Double.parseDouble(sf.getString("min_zoom")),
+          theme_max_zoom)
+        .setAttr("pmap:ne_id", sf.getString("ne_id"))
+        .setAttr("pmap:brk_a3", sf.getString("brk_a3"))
```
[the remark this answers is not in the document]
nvkelso: This might be an oversight in the documentation with the final v1.9 changes? It was added, IIRC, because it allows the cartographer the flexibility to turn off/on certain country boundary lines by ID in the style explicitly as a fail safe for legal compliance without needing a data or code change (which can take time).

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
nvkelso: For boundaries and `pmap:ne_id`, yes the intent (unlike with other layers) is to provide a fail safe during styling. This shouldn't be standard practice. Are the MVT IDs addressable in a MapLibre style?

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
nvkelso: This is a bug... since names for text labels aren't included the min_zoom here should be omitted in v3.

To your question... Yes, to enable tag continuity with the NE features. Deliberate because to the end user they don't care where the feature came from... but it should have a consistent "min_zoom" based on when the user first saw it, not when it was transitioned to another data source on the server in the "black box".

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -98,6 +196,8 @@ public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> i
     else if (zoom <= 5)
       minArea = 800 / (4096 * 4096) * (256 * 256);
     items = Area.filterArea(items, minArea);
+
+    //return FeatureMerge.mergeNearbyPolygons(items, 3.125, 3.125, 0.5, 0.5);
```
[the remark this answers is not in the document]
nvkelso: Every 1-3% counts. I was debating myself if this is useful. If we don't merge landuse then they can be stroked to separate individual "parks"... else it's just one large splotch of green. But most maps will probably not stroke them?

With natural layer it's clear cut to me that "you seen one tree, you've seen them all" haha.

CHANGELOG.md
```diff
@@ -0,0 +1,154 @@
...
+- **Release date:** 2020-10-26
+- **Credits:** [@bdon](https://github.com/bdon)
+
+- Last closed source version of the Protomaps Basemap vector tile schema
+- Add build script
+
+
+v1.0.0
+------
+- **Release date:** 2021-01-01.
```
nvkelso, line 148: @bdon Please update this with a real date

CHANGELOG.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Initial open source release as a reimplementation of the Protomaps Basemap vector tile schema as a [Planetiler](https://github.com/onthegomap/planetiler/) schema in Java.
+- Add openlayers basemap example
+- Publish new PMTiles artifact:
+    - https://r2-public.protomaps.com/protomaps-sample-datasets/protomaps-basemap-opensource-20230408.pmtiles
+
+
+v2.0.0
+------
+- **Release date:** 2020-10-26
```
nvkelso, line 139: @bdon Please check my date here

tiles/src/main/java/com/protomaps/basemap/layers/Natural.java
nvkelso: So I can remove this line entirely and the later filterArea step will still work?

tiles/src/main/java/com/protomaps/basemap/layers/PhysicalLine.java
nvkelso: So `true` not `1` (no change)?

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: Giving this a try, thanks!

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -35,123 +42,163 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
...
         feat.setAttrWithMinzoom("pmap:level", 0, 12);
       }
+
+      OsmNames.setOsmNames(feat, sf, min_zoom_names);
     }
   }
 
   @Override
   public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> items) {
+    if (zoom < 12) {
```
nvkelso, line 194: Brandon fixed a bug near here, watch for merge conflicts / pull that in now...

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
nvkelso: I prefer booleans, but OMG and MB and Tilezen and PM all do different things.

Given you like booleans, I believe this is a NOOP.

nvkelso:
Something strange is going on where I can't comment on some of your PR comments...

Said @bdon:

> Could non-building names become another POI kind?

If a building has any "business" like tags they are all handled in the POIs layer now.

If a building doesn't have any relevant POI tags, then this change removes the generic "this building is named this" tag. This is technically a regression. The way Tilezen solves this is to peel the remaining building names off as point labels, stored in the buildings layer. In Protomaps v3 those can't go in the polygon only buildings layer, so you'd either need a building_points layer, or put them as a new kind into the POIs layer. But for v3 I'm proposing to ignore this regression, as v4 has a planned fix for it and it's very edge case. (The same thing is true for address points, actually.)

> Are polygon geometries with a text property even labeled in a satisfying way using the MapLibre symbol layer?

No, polygons with names will have really odd labeling behavior on tile boundaries. This will be more noticably bad at some zooms over other zooms, and can somewhat be worked around by setting padding around the text &etc. But it's not awesome.

nvkelso:
> Do we want getSortKey here to actually work on nulls or can we just use 0 as a null value for population, population_rank, etc?

We can use `0` instead of `null` for population and population_rank.

nvkelso:
> +1 to making this based on area - think this should be expressible in pixels (planetiler considers a tile to be 256x256 https://github.com/onthegomap/planetiler/discussions/487

I find Planetiler's area representations to be super confusing. I got it to work for the POIs layer with a lot of trial and error.

But yes, this is totally doable (and once we have real "meter" units what's in Tilezen can just be copied over).

> Is this logic unique to certain situations like rows of brownstones in NYC which have similar heights? Trying to think forward to visual test cases.

Buildings layer is one of the biggest file size drivers. So this is for all types of buildings if they are attached to their neighbors or not. +1 to having brownstones in a test suite.

> This was ported from V2. There are some areal features for place=neighbourhood https://taginfo.openstreetmap.org/tags/place=neighbourhood#overview
> if it doesn't match tilezen for v3, we can remove it.

If it were me I'd recast `neighbourhood` here to `residential` landuse... but keep as `neighbourhood` for places layer.

> Is our general principle to never include names on polygonal features and instead make POIs?

Correct, that's my general principal to never include names on polygon features.


```
Pros:

Slightly smaller, but internal MVT deduplication should make this trivial
Polygon labeling on client-side has plenty of problems (finding poles of inaccessibility, crossing tiles)
Cons:

Certain use cases like clicking a feature or building a reverse geocoder from tiles will no longer work.
This is a philosophical question of if the tileset is purely for styled visualization. My gut feeling is that there is some benefit, both in practical use and debugging, of retaining names on polygons and duplicating those to POIs, with very little overhead because of MVT dedupe. Open to other opinions.
```

My 2¢: If you want revGeo tiles then make revGeo tiles.

CHANGELOG.md
bdon: that should be correct

CHANGELOG.md
bdon: updated

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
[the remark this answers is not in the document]
bdon: I'd say leave it commented out for now with a TODO - we should determine empirically if this is useful later

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
bdon: I see what you mean about `ne_id` being useful for failsafes, though it seems like referring to specific one-off features by name would be more natural. MVT IDs can addressed in the style, but in practice because I'm doing some bit twiddling the number is not going to be useful. I could be convinced to abandon MVT IDs if the storage doesn't inflate by storing all of source, ID, etc separately.

But it's important for us to have the OSM ID if possible to power debugging "where did this feature come from"

nvkelso: Note that if we switch to daylight distribution and add buildings in the IDs all get renumbered and are worthless. (I too find them great for debug.)

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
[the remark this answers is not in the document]
bdon: Sounds good to me, but if it overlaps with `ne_id` in use cases I'd prefer the one with more semantics (how often are NE IDs reassigned?)

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
bdon: I've merged that PR.

bdon:
@nvkelso I'm going to merge this now, cleanup and the few remaining tasks we can address in follow up PRs (and gives us an opportunity to flesh out the visual CI suite)

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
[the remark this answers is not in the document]
nvkelso: `ne:brk_a3` is highly stable, while the `ne:id` is mostly stable but may vary between scale set (1:10m, 1:50m, 1:110m) and major Natural Earth release versions.

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
nvkelso: Fixed in https://github.com/protomaps/basemaps/pull/79.

SEMANTIC-VERSIONING.md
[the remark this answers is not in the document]
nvkelso: Fixed in #79 via b3a953be515937e8d4f62ca19874a9d549f2f385.

Issue #53, opened 2023-07-06
link roads missing from a generated map


andrey-abstracted:
motorway_links and primary_links are present on OSM maps but missing when a freshly generated map is viewed using PMTiles viewer. 

In protomaps-basemap-opensource-20230408.pmtiles primary_links are present but motorway_links are not.

The screenshots are from the freshly generated map viewed using PMTiles viewer.

https://www.openstreetmap.org/way/454442266#map=18/-38.08778/145.32274
<img width="696" alt="Screenshot 2023-07-06 at 12 31 38 pm" src="https://github.com/protomaps/basemaps/assets/2003704/a5043cbc-4515-46c1-b3c0-aa84d0e5f693">
<img width="544" alt="Screenshot 2023-07-06 at 12 31 24 pm" src="https://github.com/protomaps/basemaps/assets/2003704/63688fbe-e9ff-472c-8b84-e4aa3e2367f5">

https://www.openstreetmap.org/way/454442266#map=15/-37.6831/144.9850
<img width="908" alt="Screenshot 2023-07-06 at 12 20 49 pm" src="https://github.com/protomaps/basemaps/assets/2003704/85a3aea7-a6b0-4af1-9b4b-8a7c35727628">
<img width="773" alt="Screenshot 2023-07-06 at 12 32 11 pm" src="https://github.com/protomaps/basemaps/assets/2003704/912e5a74-ff79-458d-8457-d1abab544308">

bdon:
Thanks, I can verify that this is an issue in specific cases where links consist of multiple ways, working on a fix.

bdon:
PR https://github.com/protomaps/basemaps/pull/57 is progress towards improving this, but the logic in LinkSimplify needs to change to handle more complex cases like the 2nd.

pperzyna:
I generated new pmtiles for planet. Happy to share to update link on website. @bdon

andrey-abstracted:
Great, thank you. On a slightly irrelevant note - would it be possible to change the zoom range for motorways from 12-15 to 11-15. That feature is quite a significant visual landmark and is still visible at lower zoom levels.

bdon:
@andrey-abstracted a major rework of both the display style and base tileset is coming soon; The best option for right now though, and in the future, is to generate your own tileset while modifying the configuration to your needs. This will likely require a small modification to both Java code and JSON styles. 

Our goal is to make this kind of adjustment easy via configuration changes and/or style parameters, but that is a medium to long-term project.

[...]

PR #57, opened 2023-07-11
stop road links from disappearing z < 12 [#53]

bdon:
@nvkelso I think this is a bug; behavior as-is makes link features vanish z<12, which isn't what we want. Changes the behavior to only add the`pmap:link` tag at 12+.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
@@ -131,7 +131,7 @@ public class Roads implements ForwardingProfile.FeatureProcessor, ForwardingProf
 
       if (sourceFeature.hasTag("highway", "motorway_link", "trunk_link", "primary_link", "secondary_link",
         "tertiary_link")) {
-        feat.setAttrWithMinzoom("pmap:link", 1, 12);
+        feat.setAttr("pmap:link", 1);
       }
 
       if (sourceFeature.hasTag("bridge", "yes")) {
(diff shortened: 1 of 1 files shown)


tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -131,7 +131,7 @@ public void processFeature(SourceFeature sourceFeature, FeatureCollector feature
 
       if (sourceFeature.hasTag("highway", "motorway_link", "trunk_link", "primary_link", "secondary_link",
         "tertiary_link")) {
-        feat.setAttr("pmap:link", 1).setZoomRange(12, 15);
+        feat.setAttrWithMinzoom("pmap:link", 1, 12);
```
nvkelso, line 134: Suggest to:

>        feat.setAttr("pmap:link", 1);

Thanks for talking me thru this today @bdon.

To whit: In the post-processing step the `linkSimplify` function classifies OSM links into either "ramps" or "connectors". Ramps (at so-called clover leaf highway interchanges) are removed at low zooms and connectors (longer links between highways that are more like the start of the highway versus a ramp) are retained. This deserves some inline comments in the functions themselves!

So I think what we want is to _always_ set `pmap:link` with `setAttr` (not `setAttrWithMinzoom`)... and then let the  postProcess function's linkSimplify do it's magic.

I got confused because Tilezen doesn't have an equivalent `linkSimplify` function so uses a brute force method that's not as sophisticated (by setting the zoom ranges) and I reimplemented that instead of upgrading it.

NOTE: I've refactored this logic in my new PR so I'll need to resolve that conflict once this is merged. No big deal.

nvkelso (changes requested):
See comment

PR #60, opened 2023-07-13
Improve LinkSimplify to work less aggressively and handle edge cases …

bdon:
…[#53]

* allow multiple ways in connectors.
* allow connectors to touch a highway at any point, not just endpoints.


bdon:
@nvkelso here's the new LinkSimplify in action:

It's not as aggressive as before, but we can add in more unit tests for new situations later!

https://github.com/protomaps/basemaps/assets/77501/6a935777-21b5-4d5f-9823-e8235a6323b2

pperzyna:
This is amazing!

PR #71, opened 2023-07-21
Refactor CountryInfos into embedded data records.

bdon:
@nvkelso


tiles/src/main/java/com/protomaps/basemap/feature/CountryInfos.java
```diff
@@ -0,0 +1,277 @@
...
+  static CountryInfo unknownInfo = new CountryInfo("UNKNOWN_COUNTRY", "XX", 8.0, 11.0);
+
+  static {
+    countryInfoByName = new HashMap<>();
+    Scanner s = new Scanner(data);
+    while (s.hasNextLine()) {
+      String line = s.nextLine();
+      String[] parts = line.split("\\|");
+      countryInfoByName.put(parts[0],
+        new CountryInfo(parts[0], parts[1], Double.parseDouble(parts[2]) - 1.0, Double.parseDouble(parts[3]) - 1.0));
```
bdon, line 266: @nvkelso what was the reason for subtracting 1 here, can we just move those into the above data? and do we need to store floats if they're only used as integers for now?

Intentionally making this flexible and type-safe so we can seamlessly add more data columns later.

nvkelso (approved):
Thanks for converting to proper data structures and adding tests!

tiles/src/main/java/com/protomaps/basemap/feature/CountryInfos.java
nvkelso: Exported min zooms for Tilezen are legacy 256 px tile size zoom values, while Planetiler prefers 512 pixel tile size zooms.

Zoom = scale so should generally be constant irrespective of tile extent (which could change to 1024 px extents in the future).

nvkelso: These are sometimes floats (and min zoom should always be thought of as floats with single tenths decimal precision (1.1 and 1.0)

PR #74, opened 2023-07-22
Refactor java variables from snake_case to camelCase; add private cto…

bdon:
…rs to utility classes.


bdon:
@nvkelso this is a pure refactor of variable names; let's consolidate on camelCase for all variables as this keeps Sonar happy and it's worth getting those linting checks down to 0.

PR #77, opened 2023-07-24
fix null pointer exceptions

bdon:
@nvkelso we should avoid nulls by using the version of the function with a default.


nvkelso (approved):

bdon:
@nvkelso `setZoomRange` always needs to be called with minZoom <= maxZoom; otherwise it will throw an assertion error. These assertions are disabled in the prod JAR, but our unit+integration tests should catch them now.

PR #79, opened 2023-07-27
Followups from PR #47 and additional polish

nvkelso:
Includes minor polish and bug fixes to #47 

- [x] Country labels at early zooms with CountryInfo functionality and tests
  - [x] Bug fixes
- [x] Region labels at early zooms with RegionInfo functionality and tests
  - [x] Bug fixes
- [x] More locality labels at early zooms
- [x] Consistent locality labels at mid zooms (San Francisco was popping in and out versus Oakland)
  - [x] Bug fixes
- [x] Remove WOF reference from SemVer statement as that's not used yet
- [x] Minor style changes / bug fixes
- [x] Add Makefile target for Kansas (helpful for debugging country and region labels in USA)
- [x] Add Makefile target for planet-xl for super well provisioned machines (and twiddle planet target)
- [x] Switch countryInfo to ISO codes (instead of fussy names) && update test
- [x] Switch regionInfo to ISO codes (instead of fussy names) && update test
- [x] Physical lines always set `pmap:level`, not just at late zooms
- [x] Places add sortKey and label grid for more consistent labeling (and smaller tiles)
- [x] POIs show smaller park labels at slightly earlier high zooms
- [x] POIs discount tall self storage buildings
- [x] POIs discount very long text names should only be shown at later zooms
- [x] POIs emphasize IATA airports
- [x] POIs de-emphasize wilderness areas
- [x] Transit include pmap:min_zoom for client side label collisions
- [x] Transit always set `pmap:level`, not just at late zooms
- [x] Don't use 110m Natural Earth data ever (use 50m instead at lowest zooms)
- [x] Boundaries from NE show 1 zoom earlier (confusing 256px versus 512px bug)
- [x] Water from NE show 1 zoom earlier (confusing 256px versus 512px bug)
- [x] OSMRef option to gather localized refs (like OSMNames) for Places layer
- [ ] Push some small zoom 14 parks in POIs layer to zoom 15 / 16 instead

In my rush out the door for vacation and then funeral I had a few uncommitted local changes that are recreated here on top of latest main.


tiles/src/main/java/com/protomaps/basemap/feature/RegionInfos.java
```diff
@@ -7,107 +7,961 @@
...
-    Yukon|CA|3.5|7.5
-    New South Wales|AU|4.6|8.1
-    Queensland|AU|4.6|8.1
-    South Australia|AU|4.6|8.1
-    Tasmania|AU|4.6|8.1
-    Victoria|AU|4.6|8.1
-    Western Australia|AU|4.6|8.1
-    Australian Capital Territory|AU|4.6|8.1
-    Northern Territory|AU|4.6|8.1
+    Australian Capital Territory|AU-ACT|3.5|7.5|Q3258
```
nvkelso, line 17: @bdon This list is getting long, so long that Java complains and suggested I put it into a file instead. I put the full list into a file, and kept a shorter version here. This should be refactored to load from the file instead. Can you submit a patch for that, please?

nvkelso:
This is ready for review, @bdon can you take a look please?

I'm still doing a few QA/QC tasks but it's ready enough for a global build now.

tiles/src/main/java/com/protomaps/basemap/feature/CountryInfos.java
```diff
@@ -253,24 +253,27 @@ public class CountryInfos {
...
     }
   }
 
-  public static CountryInfo getByName(SourceFeature sf) {
-    var name = sf.hasTag("name:en") ? sf.getString("name:en") : sf.getString("name");
-    if (countryInfoByName.containsKey(name)) {
-      return countryInfoByName.get(name);
+  public static CountryInfo getByISO(SourceFeature sf) {
+    var isoCode = sf.hasTag("ISO3166-1:alpha2") ? sf.getString("ISO3166-1:alpha2") : (
+      sf.hasTag("country_code_iso3166_1_alpha_2") ? sf.getString("country_code_iso3166_1_alpha_2") : (
```
nvkelso, line 272: Completely vexing that Geofabrik or Planetiler is changing the name of this OSM key by the time it gets here...

tiles/src/main/java/com/protomaps/basemap/feature/RegionInfos.java
```diff
@@ -7,107 +7,961 @@
...
+      regionInfoByWikidata.put(parts[4],
+        new RegionInfo(parts[0], parts[1], Double.parseDouble(parts[2]) - 1.0, Double.parseDouble(parts[3]) - 1.0, parts[4]));
+    }
+  }
+
+  public static RegionInfos.RegionInfo getByISO(SourceFeature sf) {
+    // ISO codes aren't always included in extracts, oddly, so introspect which tags are
+    //sf.tags().forEach((key, value) -> System.out.printf("%s: %s\n", key, value));
+
+    var isoCode = sf.hasTag("ISO3166-2") ? sf.getString("ISO3166-2") : (
```
nvkelso, line 947: Even though OSM has this tag, by the time we see the file via GeoFabrik and Planetiler it's gone!?

tiles/src/main/java/com/protomaps/basemap/feature/RegionInfos.java
```diff
@@ -7,107 +7,961 @@
...
+      sf.hasTag("region_code_iso3166_2") ? sf.getString("region_code_iso3166_2") : (
+        sf.hasTag("ISO3166-2") ? sf.getString("ISO3166-2") : "XX-XX")
+    );
+    if (regionInfoByISO.containsKey(isoCode)) {
+      return regionInfoByISO.get(isoCode);
     }
     return unknownInfo;
   }
+
+  public static RegionInfos.RegionInfo getByWikidata(SourceFeature sf) {
```
nvkelso, line 957: But wikidata IDs do come thru, so switch to that...

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -258,14 +264,14 @@ public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> i
 
     List<VectorTile.Feature> top64 = cities.subList(startIndex, endIndex);
 
-    // This should always be less than 64 now that a label grid is being used
+    // This should always be less than 32 now that a label grid is being used
```
nvkelso, line 267: Something is still off in this section... if I take it down to 32 from 64 some big cities still drop out in error... which suggests an error in the sort algo?

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -278,6 +278,11 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
           } else {
             minZoom = 15;
           }
+
+          // Hack for weird San Francisco university
```
nvkelso, line 282: This is a hack, but the map in SF looks really bad without it. OK to keep?

nvkelso: Probably we want to scope down more so it doesn't test every feature's name...

tiles/src/main/java/com/protomaps/basemap/feature/CountryInfos.java
bdon: Can you link an example? This shouldn't be happening at all anywhere.

tiles/src/main/java/com/protomaps/basemap/feature/RegionInfos.java
bdon: Referring to a entity like https://www.openstreetmap.org/relation/162112 right?

tiles/src/main/java/com/protomaps/basemap/feature/CountryInfos.java
nvkelso: Ah, I think I figured it out... The `{country, region}` relation has itself and a labeling node. The [relation](https://www.openstreetmap.org/relation/148838) has the expected tags (`ISO3166-1:alpha2` = `US`, but the labeling [node](https://www.openstreetmap.org/node/424317935) doesn't (but it does have `country_code_iso3166_1_alpha_2` = `US`). 

I'm used to other software (eg osm2pgsql or imposm) doing some lite processing to make them more self consistent during the extract and load phases before this transform phase in the ETL.

Here in the transform phase we probably need to walk the `rel` to see if the parent has the expected tags, else default to node's tags. **-OR-** evaluate if `wikidata` is widespread on the nodes and just use that instead (like how I migrated the Regions from ISO to wikidata).

tiles/src/main/java/com/protomaps/basemap/feature/RegionInfos.java
nvkelso: Let's defer this – it'd be better to switch to Natural Earth (at least for country labels).

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
nvkelso: With the fix:

image

Without the fix (hack) the weird SF university is shown at zoom 11 – even though no other POIs are showing. (Big parks outside the city show at this zoom, as well as airports.)

image

bdon: Is this an urban/rural divide issue? Like it looks strange in context of it being in a city, but at the same zoom level and same prominence in Davis it would look OK/

base/src/base_layers.ts
```diff
@@ -1727,9 +1727,6 @@ export function labels_layers(source: string, c: Theme): any[] {
       "source-layer": "places",
       filter: ["==", "pmap:kind", "locality"],
       layout: {
-        "symbol-sort-key": ["to-number", ["concat", ["get", "pmap:min_zoom"],
```
bdon, line 1729: This becomes redundant because the logic is the same as the sorting of the tiles internally in java.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -34,10 +34,10 @@ static int getSortKey(float min_zoom, int population_rank, long population, Stri
       .orderByInt((int) min_zoom * 10, 0, 150)
       // population_rank DESC NULLS LAST,
       //population_rank.isEmpty() ? 15 : population_rank.getAsInt()
-      .thenByInt(population_rank, 0, 15)
+      .thenByInt(population_rank, 15, 0)
```
bdon, line 37: We need to invert this: 15 population rank should be the highest priority (lowest sort key). same with population below

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -27,11 +26,12 @@ public String name() {
   private final AtomicInteger placeNumber = new AtomicInteger(0);
 
   // Evaluates place layer sort ordering of inputs into an integer for the sort-key field.
-  static int getSortKey(float min_zoom, int population_rank, long population, String name) {
+  static int getSortKey(int kind_rank, float min_zoom, int population_rank, long population, String name) {
```
bdon, line 29: Because the LabelGrid considers all places, not just cities, we need to order countries and regions before localities. So accomplish that via an integer with range 0 to 3 (0 = country, 1 = region, 2 = locality, 3 = other)

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -98,18 +94,18 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
...
         .setAttr("pmap:kind", kind)
         .setAttr("pmap:kind_detail", kindDetail)
         .setAttr("population", population)
         .setAttr("pmap:population_rank", populationRank)
         .setAttr("wikidata_id", sf.getString("wikidata"))
-        // We subtract 1 to achieve intended compilation balance vis-a-vis 256 zooms in NE and 512 zooms in Planetiler
-        .setZoomRange((int) minZoom - 1, themeMaxZoom)
-        .setBufferPixels(128)
+        .setBufferPixels(64)
+        .setPointLabelGridPixelSize(7, 64) // 64 pixels is 1/4 the tile, so a 4x4 grid
```
bdon, line 105: Let's unify all cell sizes on 64 (4x4 grid) for now - you can tweak this to taste.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: Epic, this works now!

BTW: I flipped this so it's minZoom then kindRank and added comments. All good now.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: This now works as expected. I tweaked the settings a little.

In future it may make more sense to do 3x3 grid with larger number of children... But that should vary by zoom and I don't see how to configure that in Planetiler...

base/src/base_layers.ts
nvkelso: Nice! I've applied sortKety to other label related tile layers, like POIs and Roads in subsequent comments. I'll 2x check the style layers are also updated...

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
nvkelso: Maybe? UC Davis has a large, regularly shaped polygon. This urban college is something else entirely. Perhaps more like NYU but in many different buildings 20+ blocks away from each mini campus versus having clusters of entire city blocks.

Issue #72, opened 2023-07-21
sort_rank and highroad-style layering


bdon:
In order to make correct casing ordering on complex overpasses, we may need to change Tilezen - Tangram relied on `sort_rank` with casings to overlay an arbitrary number of layers. For MapLibre GL we have `sort_key` in the style but that only works per-layer: casings needs to be drawn as separate layers.

In my experience the only solution is to have a fixed set of layers each with their own filter e.g. layers -2, -1, 0, 1, 2. @nvkelso

nvkelso:
My work around recommendation is to have -1, 0, 1, and a single new MapLibre style layer for "even higher bridges" (level = 2, 3, 4, 5, 6).

That way you avoid parsing the MVT in MapLibre "again, again, again" looking for features at those levels per style layer – when mostly no MVT data layer features are found to be eligible for the GL style layers.

Ideally the "inline" and "outline" are in the same GL style layer and could use the existing `sort_key` from Tilezen (which would need a port over to this repo). But the explosion of MapLibre style layers for each `kind` and the levels is already burdensome...

bdon:
@nvkelso I don't think there is still any solve for what you described like tilezen `sort_key` because of intersections? https://github.com/maplibre/maplibre/discussions/164#discussioncomment-4925345

So we need have 4 levels, say 5 road classes, and casings/inner strokes, 4*5*2 = 40 layers...

Unless you think we can combine all road classes into a single layer with expressions

bdon:
@wipfli what if styles weren't embedded in the properties, but you had a single layer for all classes, and the paint properties were a (very big) case statement? So a complete stack would be like

```
tunnels_casing
tunnels
ground_casing
ground
bridges_casing
bridges
bridges_2_casing
bridges
```

Instead of having `motorway`, `primary`, etc.

wipfli:
Yes the `case` approach works. A limitation is that MapLibre GL JS currently does not support data-driven line-dash arrays and some other properties like line-caps. Maybe worth talking to @zelonewolf and others from the American map style. They run into the same problem.

bdon:
So if we designed around the constraint that there are no variation in linecaps or dashes, we can get away with the total set of ~8 road layers I listed above - it would be cleaner in the number of layers, and have beastly `match`/`case` expressions - is this really better for performance than having 40 layers, or about the same?

wipfli:
regarding performance cc @msbarry

msbarry:
[...]

PR #85, opened 2023-08-13
layers attr on roads is integer [#72]

bdon:
@nvkelso 

This also starts fleshing out unit tests on layers


andrey-abstracted:
On a similar note the layers attribute on Buildings and Transit is a string. It would help to filter out underground structures if it was an integer

bdon:
Fixed by https://github.com/protomaps/basemaps/pull/100

PR #87, opened 2023-08-14
Ensure strait labels are limited to single layer (and not any water polygon exterior rings)

nvkelso:
Strait features were eligible for both physical_point and physical_line layers, producing duplicate horizontal point labels and oriented line labels in the same tile.

There should only ever be a single representation, and we should prefer the point label at this time. 

(In future we may want to generate curved lines suitable for labeling from the area, but right now we derive a label point instead.)

The result of this change keeps the blue circled node, but drops the red line.

image


nvkelso:
After:

image

image

bdon:
LGTM

PR #92, opened 2023-08-15
hawaii state routes should not be US interstates

### LINKED ISSUE #86: Road shields for Hawaii show with US interstate artwork
Instead of testing the network value for `contains("I ")`, it should be `startsWith("I ")`, else it catches networks like `HI XXXX` in error. My bad as a newbie to Java but easy to fix.

For example:

- https://www.openstreetmap.org/way/44750946#map=14/20.8742/-156.4637

image

https://github.com/protomaps/basemaps/blob/main/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java#L40-L45

```
if (shieldText.contains("US ")) {
          shieldText = shieldText.replaceAll("US ", "");
          networkVal = "US:US";
        } else if (shieldText.contains("I ")) {
          shieldText = shieldText.replaceAll("I ", "");
          networkVal = "US:I";
        } else {
          // This should be replaced by walking the way's relations (which reliably set network)
          networkVal = "other";
        }
```

nvkelso:
Basic Java string manipulation change.

Fixes #86

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
@@ -36,10 +36,10 @@ public class Roads implements ForwardingProfile.FeatureProcessor, ForwardingProf
       String networkVal = sf.getString("network");
       shieldText = (shieldText == null ? null : shieldText.split(";")[0]);
       if (shieldText != null) {
-        if (shieldText.contains("US ")) {
+        if (shieldText.startsWith("US ")) {
           shieldText = shieldText.replaceAll("US ", "");
           networkVal = "US:US";
-        } else if (shieldText.contains("I ")) {
+        } else if (shieldText.startsWith("I ")) {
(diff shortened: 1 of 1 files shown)


bdon (approved):

bdon:
@nvkelso can you link some OSM examples so I can flesh out the test suite?

nvkelso:
@bdon from the linked issue:

- https://www.openstreetmap.org/way/44750946#map=14/20.8742/-156.4637

With the following tags:

tag | value
--- | ------
highway | tertiary
lanes | 2
name | South Wakea Avenue
ref | HI-3920

Where the ref should export as `HI3920` (not `HI-3920`) and the network should be `other` (not `US:I`).

NOTE: A later change I'll propose will strip `HI` off the ref, and later more set the network to `US:HI`). So the test will be 3x useful :)

bdon:
@nvkelso  PTAL #114 - can encapsulate logic and test cases for HI-3920 -> HI3920

Issue #91, opened 2023-08-15
Places layer localities reset min_zoom to 8 at zoom 7


nvkelso:
The earlier (and smaller value) zooms from Natural Earth from zooms 0-6 are lost on zoom in. 

For **New Delhi** in India this means the locality which had a `pmap:min_zoom` of `2.0` reset to `8.0`, which compares unfavorably to **Delhi** the `region` that has a `pmap:min_zoom` of 4.6. 

As a result the **New Delhi** `locality` is collided in error by the **Delhi** `region` label for several zooms from 7 until room opens up.

A solution is to do a data join in the **places** layer for `locality` features between NE and OSM, and prefer the NE `min_zoom` values.

<img width="1235" alt="image" src="https://github.com/protomaps/basemaps/assets/853051/71d8468a-e44b-4abc-8f11-fba0a8a02c29">

PR #93, opened 2023-08-15
add NE <> OSM data join for places layer

nvkelso:
Fixes https://github.com/protomaps/basemaps/issues/91.


tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -25,6 +29,17 @@ public String name() {
 
   private final AtomicInteger placeNumber = new AtomicInteger(0);
 
+  // To have consistent min_zoom and other attr between Natural Earth (low zooms) and OpenStreetMap (high zooms)
+  // we need to store the NE places with a spatial indexes for later joins with OSM
+  private PointIndex<NaturalEarthPlace> ne_populated_places = PointIndex.create();
```
nvkelso, line 34: After creation inserts are blocked?

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -52,6 +72,22 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
     if (!sourceLayer.equals("ne_10m_populated_places")) {
       return;
     }
+
+    // Setup for high zoom content
```
nvkelso, line 76: Can this be moved out of per tile processing into project startup?

bdon: I think it's a good idea, you said only 8000 places right? I don't want to slow down the small job tests by loading in the whole thing every time

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -25,6 +29,17 @@ public String name() {
...
 
+  // To have consistent min_zoom and other attr between Natural Earth (low zooms) and OpenStreetMap (high zooms)
+  // we need to store the NE places with a spatial indexes for later joins with OSM
+  private PointIndex<NaturalEarthPlace> ne_populated_places = PointIndex.create();
+
+  // Data structure for any Natural Earth place (eg for data joins between NE and OSM)
+  private record NaturalEarthPlace(String name, String wikidataId, float minZoom, int populationRank) {}
+
+  // Search window size for NE <> OSM data joins
+  private static final double LOCALITY_JOIN_DISTANCE = GeoUtils.metersToPixelAtEquator(0, 50_000) / 256d;
```
bdon, line 40: from what I read below, this is only necessary in 2% of the cases that don't have a Wikidata ID?

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: Verified this works in a build with no errors... as long as NE is added to Planetiler profile before OSM.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: The current config doesn't seem to noticeably affect build time so it's probably OK as is?

But it's better architecture to only load it once on project init. We can file a followup issue for that?

nvkelso:
Before low zoom:

image

Before high zoom:

image

After high zoom:

image

nvkelso:
@bdon back at ya

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: Followup in https://github.com/protomaps/basemaps/issues/108.

PR #96, opened 2023-08-17
throttle back early airports

nvkelso:
While the earlier logic did promote labels for international airports to earlier zooms, overall all airport labels were showing up too early. This change caps them to zoom 10 for international airports and zoom 12 for regional airports.


nvkelso:
This could also be accomplished with a separate zoom - area grading, but this works for now.

Before:

image

After: 

image

Before:

image

After:

image

PR #98, opened 2023-08-17
Adjust places layer thinning

### LINKED ISSUE #89: Some state (region) label points are dropped at some early zooms
This is especially bad at zooms 2 and 3 in the USA.

Likely part of the label thinning for `places` layer. Perhaps that should be limited to `locality` and `neighbourhood` features only?

Zoom 2:
image

Zoom 3:
image

nvkelso:
Fixes #89 

- [ ] More sister city pairs (SF and Oakland)
- [ ] More region labels


nvkelso:
Red dots are new, blue dots are from before/after

image

image

image

image

Perhaps too many at zoom 9?

image

bdon:
LGTM

Issue #117, opened 2023-09-06
Qrank related build problem


nvkelso:
I'm seeing a few of these building from main after the QRank PR was merged. We probably need to gaurd against `;` delim OSM values?

```0:20:53 ERR [osm_pass2:process] - Error processing OSM Way 785296098
java.lang.NumberFormatException: For input string: "49514848;Q49321752"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Long.parseLong(Long.java:708)
	at java.base/java.lang.Long.parseLong(Long.java:831)
	at com.protomaps.basemap.feature.QrankDb.get(QrankDb.java:30)
	at com.protomaps.basemap.layers.Pois.processFeature(Pois.java:63)
	at com.onthegomap.planetiler.ForwardingProfile.processFeature(ForwardingProfile.java:116)
	at com.onthegomap.planetiler.reader.osm.OsmReader.render(OsmReader.java:468)
	at com.onthegomap.planetiler.reader.osm.OsmReader.lambda$pass2$7(OsmReader.java:360)
	at com.onthegomap.planetiler.worker.WorkerPipeline$Builder.lambda$addWorker$0(WorkerPipeline.java:252)
	at com.onthegomap.planetiler.worker.Worker.lambda$new$0(Worker.java:41)
	at com.onthegomap.planetiler.worker.Worker.lambda$new$1(Worker.java:68)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1623)
```

nvkelso:
Testing just in SF bay area there was 2nd example (but same type of bug). For the whole world there are probably more examples.

```
0:19:55 ERR [osm_pass2:process] - Error processing OSM Way 461759932
java.lang.NumberFormatException: For input string: "49553887;Q49553890"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Long.parseLong(Long.java:708)
	at java.base/java.lang.Long.parseLong(Long.java:831)
[...]

PR #118, opened 2023-09-07
handle multiple or misformatted wikidata id [#117]

bdon:
@nvkelso ptal

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/feature/QrankDb.java
+++ b/tiles/src/main/java/com/protomaps/basemap/feature/QrankDb.java
@@ -26,9 +26,16 @@ public final class QrankDb {
     return this.db.get(wikidataId);
   }
 
-  public long get(String wikidataId) {
-    long id = Long.parseLong(wikidataId.substring(1));
-    return this.get(id);
+  public long get(String osmValue) {
+    try {
+      if (osmValue.contains(";")) {
+        osmValue = osmValue.split(";")[0];
+      }
+      long id = Long.parseLong(osmValue.substring(1));
+      return this.get(id);
+    } catch (NumberFormatException e) {
+      return 0;
+    }
   }
 
   public static QrankDb fromCsv(Path csvPath) throws IOException {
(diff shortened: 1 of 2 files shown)


tiles/src/main/java/com/protomaps/basemap/feature/QrankDb.java
```diff
@@ -26,9 +26,16 @@ public long get(long wikidataId) {
...
 
-  public long get(String wikidataId) {
-    long id = Long.parseLong(wikidataId.substring(1));
-    return this.get(id);
+  public long get(String osmValue) {
+    try {
+      if (osmValue.contains(";")) {
+        osmValue = osmValue.split(";")[0];
+      }
+      long id = Long.parseLong(osmValue.substring(1));
```
nvkelso, line 34: Wikidata IDs usually start with a `Q`? In the linked issue the 0 index position values lacks the Q, while the 1st index position has it. Might need special logic around this (eg to add or strip Q prefix and treat like an int?)

(repeating here in PR instead of earlier comment on a commit)

tiles/src/test/java/com/protomaps/basemap/feature/QrankDbTest.java
```diff
@@ -26,4 +26,13 @@ void testLookup() {
     assertEquals(1234, db.get("Q1"));
     assertEquals(0, db.get("Q2"));
   }
+
+  @Test
+  void testLookupMultiple() {
+    var hashMap = new LongLongHashMap();
+    hashMap.put(1, 1234);
+    var db = new QrankDb(hashMap);
+    assertEquals(1234, db.get("Q1;Q2"));
```
nvkelso, line 35: Also add test for `"1;Q2"` as well to match the example provided in the bug report?

bdon: @nvkelso the bug report is showing the exception when this runs

`Long.parseLong(osmValue.substring(1));`

the substring(1) is stripping off the Q

tiles/src/main/java/com/protomaps/basemap/feature/QrankDb.java
bdon: Version 5 here is the actual OSM entity: https://www.openstreetmap.org/way/461759932/history which has Qs in both places

PR #128, opened 2023-09-18
Fix minzoom keys in base_layers

bdon:
@nvkelso it turns out having the wrong key `min_zoom` instead of `minzoom` is not only ineffective, it also silently breaks smooth style diffing! Weird. (TODO: figure out a better way to catch this with TypeScript)


nvkelso:
Ah, this quirk of MapLibre (inherited from Mapbox) styling always gets me. Thanks for the fix!

bdon:
https://github.com/protomaps/basemaps/pull/130 added type safety to catch this ;)

PR #161, opened 2023-10-25
Use java 21

msbarry:
I'm upgrading planetiler soon to require java 21 so am updating a few projects that use planetiler ahead of that.  This also unlocks using some [new features](https://advancedweb.hu/a-categorized-list-of-all-java-and-jvm-features-since-jdk-8-to-21/) that should make java little more ergonomic.


msbarry:
cc/ @nvkelso @bdon you'll want to update your local environment to use java 21 ahead of this!

nvkelso:
Same, thanks for the tip and keeping things current :)

PR #109, opened 2023-08-25
wip multiexpression

bdon:
@nvkelso this is a WIP on a refactor of landuse into an expression

PR #164, opened 2023-10-26
Adds bus stops to POIs

### LINKED ISSUE #163: Bus stops missing in POIs
The tileset does not include bus stops. It would be nice to have them.

eikes:
This PR fixes #163 by adding `highway=bus_stop` to the POIs

See https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_stop

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -51,6 +51,7 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
       sf.hasTag("leisure") ||
       sf.hasTag("natural", "beach") ||
       sf.hasTag("railway", "station") ||
+      sf.hasTag("highway", "bus_stop") ||
       sf.hasTag("shop") ||
       sf.hasTag("tourism") &&
         (!sf.hasTag("historic", "district")))) {
@@ -134,6 +135,8 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
           kind = sf.getString("natural");
         } else if (sf.hasTag("railway")) {
           kind = sf.getString("railway");
+        } else if (sf.hasTag("highway")) {
+          kind = sf.getString("highway");
         } else if (sf.hasTag("shop")) {
           kind = sf.getString("shop");
         } else if (sf.hasTag("tourism")) {
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -209,6 +209,10 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
         kindDetail = sf.getString("sport");
       }
 
+      if (sf.hasTag("highway", "bus_stop")) {
+        minZoom = 16;
+      }
+
       // try first for polygon -> point representations
       if (sf.canBePolygon() && sf.hasTag("name") && sf.getString("name") != null) {
         Double wayArea = 0.0;
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -114,6 +114,8 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
       } else if (sf.hasTag("tourism", "attraction", "camp_site", "hotel")) {
         kind = sf.getString("tourism");
         minZoom = 15;
+      } else if (sf.hasTag("highway", "bus_stop")) {
+        minZoom = 18;
       } else {
         // Avoid problem of too many "other" kinds
         // All these will default to min_zoom of 15
@@ -207,10 +209,6 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
         kindDetail = sf.getString("religion");
       } else if (sf.hasTag("sport")) {
         kindDetail = sf.getString("sport");
-      }
-
-      if (sf.hasTag("highway", "bus_stop")) {
-        minZoom = 16;
       }
 
       // try first for polygon -> point representations


nvkelso (changes requested):
Changes look good so, thanks for the PR!

Please also set a high min_zoom on the new features like 18 so they are only in the max zoom tiles (zoomed into street level). That happens towards bottom of this same file.

eikes:
I set the minZoom to 16 as suggested by @bdon in the issue #163

nvkelso:
Looking at the Tilezen source code (ultimate source of truth versus docs), zoom 18 is more appropriate. Otherwise they really crowd the map.

- https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1974-L1993

I'll update that here and merge, thanks!

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -134,6 +137,8 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
           kind = sf.getString("natural");
         } else if (sf.hasTag("railway")) {
           kind = sf.getString("railway");
+        } else if (sf.hasTag("highway")) {
```
nvkelso, line 140: This isn't strictly necessary (since we are just doing highway=bus_stop now), but good to include as a way to future proof later tag value additions.

eikes:
@nvkelso That makes sense when the tilezen style is the desired result. The current OSM Carto style uses 16 / 17 for the bus_stops: https://github.com/gravitystorm/openstreetmap-carto/blob/4abe70d9fea9eb58927390cefe50faf14dc19282/style/amenity-points.mss#L149-L161

My question is: should it not be left to the style designer to chose what should be visible at what level? I know very little about protomaps so please excuse my lack of knowledge: My understanding is that the code we are discussing here creates the protomaps vector tiles. When displaying them you still have the ability to chose what to show (and even at which level). But you cannot put everything at every level, so you have to pre-select it, right? If you decide now, that bus stops should only be visible at zoom level 18 or greater, I, as a style designer, cannot show the bus_stops earlier, right? 

I work a lot with public transport and here in Europe that's a big thing, so I would appreciate it if the bus stops would be included earlier. But if you want to stick with level the tilezen style, than that's ok too! 

Last words: I stumbled upon the protomaps project and was able to download a small area and use maplibre gl js to render it. Works like a charm, lovely developer experience. Will definitely migrate from mapbox to self-hosted. I understand this is a ton of work. So thank you!

nvkelso:
Thanks for the background, now I understand your use case. Aside: a long time ago I designing Apple's Transit map mode.

Overall in Protomaps basemap repo we're balancing small default tiles and somewhat opinionated base maps styles with larger tile sizes and more styling flexibility. 

In your case, because of how protomaps works with max zoom 15 tiles, the bus stops will be available starting at that zoom even if the default style won't show them until later. 

When you style them in your overlay (or in a style override), do so by selecting their kind value in your style layer and it should all work out.

eis-ioki:
Thank you @bdon and @nvkelso for your explanations! I will try to learn more about the tile layers of protomaps. It seems like there are some things I don't understand just yet... 😄

eikes:
Thank you for your explanations @bdon and @nvkelso! I will try to learn more about the protomaps tile layers. It seems like there are some things I don't understand just yet...

bdon:
> SonarCloud is complaining, I'll need to circle back later to debug...

don't worry about this for now, will improve once we have #109 applied to most layers

eikes:
I was taking a look at the xrays of the most recent map builds and didn't see the bus_stops there. So I wanted to ask if something is wrong with this PR or if the Planetiler profile is deployed manually?

bdon:
@eikes thanks for checking, I'll take a look at the missing bus stops today

bdon:
Screenshot 2023-11-04 at 16 54 22

will open a PR with 1 line change

eikes:
Is that a custom build? Because here I can't see this bus stop:

https://protomaps.github.io/PMTiles/?url=https%3A%2F%2Fbuild.protomaps.com%2F20231105.pmtiles#map=18.09/37.762231/-122.415139

bdon:
@eikes yes, it's a local build on my dev machine for just one small area - you can run the JAR on any area from http://app.protomaps.com/downloads/osm

eikes:
After a lot of mucking around I finally got it to work:

https://maps.eike.se/#17.45/52.520419/13.404826

Here is what I needed to do:

```diff
diff --git a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
index 173e293..13d46e7 100644
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -115,7 +115,8 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
         kind = sf.getString("tourism");
         minZoom = 15;
       } else if (sf.hasTag("highway", "bus_stop")) {
-        minZoom = 18;
+        kind = sf.getString("highway");
+        minZoom = 15;
       } else {
         // Avoid problem of too many "other" kinds
         // All these will default to min_zoom of 15
@@ -137,8 +138,6 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
           kind = sf.getString("natural");
         } else if (sf.hasTag("railway")) {
           kind = sf.getString("railway");
-        } else if (sf.hasTag("highway")) {
-          kind = sf.getString("highway");
         } else if (sf.hasTag("shop")) {
           kind = sf.getString("shop");
         } else if (sf.hasTag("tourism")) {
```

Any minZoom higher than 15 will not be created in the tiles. And I have to set the kind where I set the minZoom, the other `else if` will never be reached...

I will open another PR to fix that.

bdon:
@eikes very nice! thanks for working through that. The POIs code needs a big refactor for readability to make these changes easier which will follow along #109 . I will prioritize getting those merged to smooth things out.

PR #166, opened 2023-10-26
Use station tag value in pmap:kind_detail and include tram stops in POIs

### LINKED ISSUE #165: Differentiate train station types
**Description**
The train station information should include what kind of train station it is, for example: subway, light_rails or regular train...

eikes:
Use station tag value in pmap:kind_detail to differentiate subway / light_rail / train stations

Include railway=tram_stop in POIs

This PR fixes #165 

See:
https://wiki.openstreetmap.org/wiki/Key:station
https://wiki.openstreetmap.org/wiki/Tag:railway%3Dtram_stop

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -215,7 +215,7 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
         kindDetail = sf.getString("station");
       }
       
-      if (sf.hasTag("railway", "station") && 
+      if (sf.hasTag("railway", "station") &&
         (!sf.hasTag("station") || sf.hasTag("station", "train", "light_rail", "subway"))) {
         minZoom = 10;
       } else if (sf.hasTag("railway", "halt", "tram_stop")) {
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -52,6 +52,7 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
       sf.hasTag("natural", "beach") ||
       sf.hasTag("railway", "station") ||
       sf.hasTag("railway", "tram_stop") ||
+      sf.hasTag("railway", "halt") ||
       sf.hasTag("shop") ||
       sf.hasTag("tourism") &&
         (!sf.hasTag("historic", "district")))) {
@@ -211,7 +212,7 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
       
       if (sf.hasTag("railway", "station") && (!sf.hasTag("station") || sf.hasTag("station", "train"))) {
         minZoom = 13;
-      if (sf.hasTag("station", "subway", "light_rail")) {
+      } else if (sf.hasTag("station", "subway", "light_rail") || sf.hasTag("railway", "halt")) {
         minZoom = 14;
       } else if (sf.hasTag("railway", "tram_stop")) {
         minZoom = 16;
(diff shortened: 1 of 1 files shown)


nvkelso (commented):
Should min_zoom carry by the new station kind_detail value?

Same comment as other Pr, let's set an explicate min_zoom on new tram_stop, too. Probably zoom 15 or 16?

eikes:
> Should min_zoom carry by the new station kind_detail value?
> 
> Same comment as other Pr, let's set an explicate min_zoom on new tram_stop, too. Probably zoom 15 or 16?

To match bus stops, I set the tram_stop to zoom 16. Regular train stations should be as important as visible as libraries right? So I set it to 13 as well. Subway and light rail at 14 seemed appropriate. WDYT @nvkelso ?

eikes:
I also added `railway: halt`

bdon:
Do we need to use a different tagging scheme once we have a single POI that is part of multiple transit modes?

https://tilezen.readthedocs.io/en/latest/layers/#poi-properties-only-on-kindstation

eikes:
These transit networks are distinct. I don't think that these nodes exist. As far as I can tell, these stations are mapped as distinct nodes as well, even when they are in the same building.

eikes:
I changed the zoom for the `train` (implicit if absent) and `light_rail` stations to minzoom 10 match the tilezen definitions:

https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1948-L1953

(I also set  `subway` to the same, because it is not explicitly set in the tilezen definition.)

I set `tram_stop` and `halt` to minzoom 16:

https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1958-L1965

eikes:
I finally fixed all the code formatting issues!

bdon:
Thanks for fixing the formatting!

Ultimately the nature of this tag is to power a different icon, and the one we have in Tilezen looks like this:

station

This is used indiscriminately for light rail, subway, tram.

Committing to a certain tag means we need to stick with that tag going forward for all styles. @eikes can you give more background on how your style plans to use this tag?

bdon:
Oh , we also have `kind=tram_stop` 

https://tilezen.readthedocs.io/en/latest/layers/

tram_stop

@nvkelso open to guidance on whether we want to stick with pure Tilezen or pursue a different tagging scheme, previous linked issue: https://github.com/tilezen/vector-datasource/issues/532

eikes:
I also wanted to mention, that there are in fact nodes which have two conflicting kinds at the same time: 

https://overpass-turbo.eu/s/1D92

nvkelso:
👋 Sorry for the delay, I was traveling. 

Expanding the [docs](https://tilezen.readthedocs.io/en/latest/layers/#poi-properties-only-on-kindstation) that @bdon linked to: 

In Tilezen, any `station` POI can have a set of `is_*` boolean flags indicating whether the station has any routes of the given type. These are: `is_train`, `is_subway`, `is_light_rail`, `is_tram`, corresponding to the above `*_routes`. This is provided as a convenience for styling. 

The booleans are calculated by walking the transit relations that go thru the station node (reading properties on the node is not reliable), and would be new logic to add. But walking relations in Planetiler is "free", so "easy"? ;)

This is because for most general usage maps it's just "something train, something on rails", while for a transit map style you'd want custom icons (and even grouped icons). 

For the cases where the type of station is ambiguous... `station` should imply some sort of "rails" station, and the aerialway station should get a distinct new kind value (eg `aerialway_station` since it's icon should be different).

eikes:
That's true! You can't really put two icons on the same node anyways. But as a designer of a map I might be more interested in on kind of node than the other.

bdon:
Sorry, I have to close this PR for now - please keep the issue open so we can revisit it! We are focusing on achieving parity with the Tilezen project in 2024, and because this feature expands beyond the tagging scheme of Tilezen it's difficult to incorporate into the design right now. At a future date we may build beyond the Tilezen schema or offer a configuration-driven way to add extra tags.

PR #176, opened 2023-11-05
bus_stop zoom off by 1 [#163]

bdon:
@nvkelso @eikes 

We should document how this works at https://docs.protomaps.com/basemaps/layers 

notably, we need to be consistent about the mismatch between 256px NE/Leaflet zooms and MapLibre zooms. when we want bus stops to display at Z18 (MapLibre) we encode min_zoom = 17 in the tiles?


nvkelso:
Yes, we could straighten this out.

In the POIs layer we're setting the kind min_zooms to the 512px values (so -1 from what is in Tilezen, to then increase them on export)... also so setting the zoomRange is more straight forward. 

> .setAttr("pmap:min_zoom", minZoom + 1)
...snip...
> .setZoomRange(Math.min(minZoom, 15), 15)

I think the other layers also follow this pattern?

Issue #119, opened 2023-09-11
NE boundaries get "doubled" (visible while zooming in)


bdon:

bdon:
@nvkelso this boils down to duplication of NE boundaries from zooms 0-3 somehow

https://github.com/protomaps/basemaps/blob/main/tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java#L33

Can we remove the 50m dataset in favor of using the 10m dataset only?

nvkelso:
I have an idea, let me submit a PR...

PR #177, opened 2023-11-09
Only use ne_10m scale for boundaries [#119]

bdon:
* We do our own generalization in code, so we don't need the generalized tables.

Fixes duplicate boundaries where multiple ne versions don't match up @nvkelso 

Before: (exaggerated overzoom)

Screenshot 2023-11-10 at 07 32 22

After:

Screenshot 2023-11-10 at 07 33 10


nvkelso:
Thanks!

Issue #59, opened 2023-07-12
enhance metadata with more version information


bdon:
examples: NE version, git SHA of basemaps repo

PR #178, opened 2023-11-10
Move tileset version into Basemap.java [#59]

bdon:
Also remove dead code related to old CI system for limited areas.

@nvkelso this moves `3.0.0-pre.1` into the Java code so it is fetchable as `version` in the pmtiles metadata.

This will let us list the version on https://maps.protomaps.com/builds/


nvkelso:
Brilliant!

PR #180, opened 2023-11-11
Bus stops bridges

bdon:
Fixes display of bridges shown in #67 

Assigns correct `pmap:kind` to bus stops in #163 

@wipfli

Screenshot 2023-11-11 at 08 36 21

Screenshot 2023-11-11 at 08 36 16

PR #182, opened 2023-11-12
don't require man_made=bridge outlines have area=yes [#180]

bdon:
See #180 

Refer to https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge : 

> The tag [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=bridge is used to tag a bridge defined by the outline of the bridge.

@wipfli @nvkelso

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
@@ -21,7 +21,7 @@ public class Landuse implements ForwardingProfile.FeatureProcessor, ForwardingPr
       sf.hasTag("landuse", "recreation_ground", "industrial", "brownfield", "railway", "cemetery", "commercial",
         "grass", "orchard", "farmland", "farmyard", "residential", "military") ||
       sf.hasTag("leisure", "park", "garden", "golf_course", "dog_park", "playground", "pitch", "nature_reserve") ||
-      sf.hasTag("man_made", "pier") || sf.hasTag("man_made", "bridge") ||
+      sf.hasTag("man_made", "pier", "bridge") ||
       sf.hasTag("natural", "beach") ||
       // TODO: (nvkelso 20230622) This use of the place tag here is dubious, though paired with "residential"
       sf.hasTag("place", "neighbourhood") ||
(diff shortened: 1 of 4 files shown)


wipfli:
Looks good to me

tiles/src/main/java/com/protomaps/basemap/layers/Landuse.java
```diff
@@ -21,14 +21,14 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
       sf.hasTag("landuse", "recreation_ground", "industrial", "brownfield", "railway", "cemetery", "commercial",
         "grass", "orchard", "farmland", "farmyard", "residential", "military") ||
       sf.hasTag("leisure", "park", "garden", "golf_course", "dog_park", "playground", "pitch", "nature_reserve") ||
-      sf.hasTag("man_made", "pier") ||
+      sf.hasTag("man_made", "pier") || sf.hasTag("man_made", "bridge") ||
```
nvkelso, line 24: Nit, since these are both the same tag key, rewrite asa list of values, like:

>       sf.hasTag("man_made", "pier", "bridge") ||

nvkelso: Agree that in the case of `bridge` it's assumed to be a polygon per the [wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge) and in practice.

nvkelso: Does this need a style change, too? Something like `id: "landuse_pedestrian",` either duplicating that style layer or adding `bridge` to it's the filter?

bdon: It's covered here: https://github.com/protomaps/basemaps/blob/main/styles/src/base_layers.ts#L286

nvkelso: Can you include a before/after screenshot showing the data change in the style? 

(I think that style layer filter needs a logic change...)

bdon: Will provide visual test link when the 3.0.1 build completes tomorrow... think the style with `pmap:kind=pedestrian` is good enough

bdon:
@nvkelso visual test output comparing tileset builds:

Screenshot 2024-01-03 at 16 31 49

https://maps.protomaps.com/visualtests/?leftTiles=https%3A%2F%2Fbuild.protomaps.com%2F20240101.pmtiles&name=zurich-bridges

nvkelso:
Great bug fix!

PR #185, opened 2023-11-22
Closes #184 (Adding a dockerfile & instructions on the README)

### LINKED ISSUE #184: Generate tiles with Docker
**Description**
I (and I assume probably others as well) need the Planetiler build for an area of my interest, but want to avoid installing Java locally. An ready-to-go, dockerized setup would save much time from my workflow (I 've had problems setting up multiple Java versions in the past)

lymperis-e:
I have added a Dockerfile and a relevant section in the README, so that the tile egneration workload can be executed in a Docker container.


bdon:
See https://github.com/protomaps/basemaps/pull/269 which may be simpler

Issue #183, opened 2023-11-17
Too large font for places_region


Edefritz:
**Description**
As far as I understand, the places_region layer renders names of states/regions. While this works well for US states, it creates comparatively huge labels for regions in other parts of the world. 
The region labels are larger than the country labels and somehow make the map appear cluttered.

I think this is both, a data and a style "issue". 

There also seems to be a bug as US state labels seem to disappear after z7 but the state labels in other parts of the world remain rendered:
https://maps.protomaps.com/#map=7.52/42.67/-107.473&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231120.pmtiles
https://maps.protomaps.com/#map=9.64/52.7965/13.1533&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231120.pmtiles

@nvkelso [left a comment](https://github.com/protomaps/basemaps/blob/603091aff3d54499f8e60c33030b8616fe87cf67/tiles/src/main/java/com/protomaps/basemap/layers/Places.java#L192) writing that eventually the source for this layer should transition to Natural Earth data anyway. But in the meantime I would suggest to reduce the font size so the map remains

**Location**
Looks ok:
https://maps.protomaps.com/#map=5.28/37.241/-83.096&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231120.pmtiles
Not so much:
https://maps.protomaps.com/#map=5.28/52.416/12.78&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231120.pmtiles

**Screenshots**
<img width="1715" alt="Screenshot 2023-11-17 at 15 04 17" src="https://github.com/protomaps/basemaps/assets/1841670/d6955196-de57-42aa-abda-536a25629496">

<img width="1123" alt="Screenshot 2023-11-17 at 14 55 22" src="https://github.com/protomaps/basemaps/assets/1841670/b3f5a9cb-633d-426d-8251-aa94b4ece2ba">

I'm not sure how to best address this. I feel changing the style to something like this already looks much tidier:
```javascript
    {
      id: "places_region",
      type: "symbol",
      source: source,
      "source-layer": "places",
      filter: ["==", "pmap:kind", "region"],
      layout: {
        "symbol-sort-key": ["get", "pmap:min_zoom"],
        "text-field": [
          "step",
          ["zoom"],
          ["get", "name:short"],
          6, // only show full name from zoom 6
          ["get", "name"],
        ],
[...]

PR #192, opened 2023-12-19
Address issue with state labels being too large

Edefritz:
Address issues described here: https://github.com/protomaps/basemaps/issues/183

- only show full state label from zoom 6+
- max font size for state labels capped at 16


bdon:
Thanks!

Issue #189, opened 2023-12-09
Missing boundaries in various places `pmap:min_admin_level	4`


makella:
**Location**
kenya: https://maps.protomaps.com/#map=6.29/0.017/37.603&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231225.pmtiles

columbia:
https://maps.protomaps.com/#map=7/4.518/-71.805&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231225.pmtiles

brazil:
https://maps.protomaps.com/#map=6.83/-15.131/-46.307&theme=light&renderer=maplibregl&tiles=https://build.protomaps.com/20231225.pmtiles

assuming this is happening in other places as well

**Screenshots**
when zooming in, the boundaries become disjointed as large parts disappear for example here in brazil you see them at z5 and then they disappear if not totally partially
<img width="744" alt="Screen Shot 2023-12-08 at 6 25 40 PM" src="https://github.com/protomaps/basemaps/assets/1566273/d062ad15-0231-482f-bbac-ca194a1b93c2">

<img width="1025" alt="Screen Shot 2023-12-08 at 6 25 53 PM" src="https://github.com/protomaps/basemaps/assets/1566273/d17ef0e3-f422-43ab-8e62-f8170c8af31a">

**Required information**
 - Renderer version [e.g. MapLibre GL JS 3.3.0]
 - Tileset version and source
 - Style JSON or NPM version

bdon:
@nvkelso for Kenya, the NE release we're using seems to have the [pre-2010 provinces of Kenya](https://en.wikipedia.org/wiki/Provinces_of_Kenya) instead of [post-2010 counties](https://en.wikipedia.org/wiki/Counties_of_Kenya)

bdon:
@nvkelso fundamental issue here is that Brazil, Colombia and Kenya use [`admin_level=3` and `admin_level=5` in OSM](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative#10_admin_level_values_for_specific_countries)

[Tilezen boundaries](https://tilezen.readthedocs.io/en/latest/layers/#boundaries) don't specify where 3 and 5 should go. We should focus our tileset on display needs and not an arbitrary database - easiest fix is to simply allow 3 and 5 for `pmap:min_admin_level`

bdon:
This should be resolved in tiles 3.1.0:

[...]

PR #202, opened 2024-01-05
Boundaries admin level 3 5

bdon:
See #189 for screenshots


tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -199,6 +199,13 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
             // Natural Earth is used for low zooms (for compilation and tile size reasons)
             themeMinZoom = 6;
           }
+          // used in Colombia, Brazil, Kenya (historical)
```
nvkelso, line 202: Beware: In Brazil there are several "statistical" groupings of actual administrative regions. Some of these are in conventional usage, some are not. I'm also wary of adding historical boundary lines. However, since the map looks broken without this change we should make it... but worth some more research...

PR #212, opened 2024-01-24
Closes #204 and adds POIs for landuse=village_green, laduse=allotment…

### LINKED ISSUE #204: Some landuse elements missing: village_green, allotments, playground
**Description**
OSM tagged areas with `landuse=village_green` and `landuse=allotments` are missing in the tileset. Same is true for polygons with `leisure=playground`. I think they should all be included in layer "landuse" with

- kind=village_green
- kind=allotments
- kind=playground

**Location**
https://www.openstreetmap.org/#map=18/51.50497/7.31888

**Screenshots**
grafik
vs
grafik

I can provide a PR, if that helps :-)

lenalebt:
…s, leisure=playground

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
@@ -39,7 +39,8 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
       sf.hasTag("boundary", "national_park", "protected_area") ||
       sf.hasTag("craft") ||
       sf.hasTag("historic") ||
-      sf.hasTag("landuse", "cemetery", "recreation_ground", "winter_sports", "quarry", "park", "forest", "military") ||
+      sf.hasTag("landuse", "cemetery", "recreation_ground", "winter_sports", "quarry", "park", "forest", "military",
+        "village_green", "allotments") ||
       sf.hasTag("leisure") ||
       sf.hasTag("natural", "beach") ||
       sf.hasTag("railway", "station") ||
@@ -323,7 +324,8 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
         } else if (kind.equals("forest") ||
           kind.equals("park") ||
           kind.equals("protected_area") ||
-          kind.equals("nature_reserve")) {
+          kind.equals("nature_reserve") ||
+          kind.equals("village_green")) {
           if (wayArea > 10000) {
             minZoom = 7;
           } else if (wayArea > 4000) {
@@ -368,6 +370,14 @@ public class Pois implements ForwardingProfile.FeatureProcessor, ForwardingProfi
             minZoom = 16;
           }
           // Typically for "building" derived label placements for shops and other businesses
+        } else if (kind.equals("allotments")) {
+          if (wayArea > 0.01) {
+            minZoom = 15;
+          } else {
+            minZoom = 16;
+          }
+        } else if (kind.equals("playground")) {
+          minZoom = 17;
         } else {
           if (wayArea > 10) {
             minZoom = 11;
--- a/tiles/src/test/java/com/protomaps/basemap/layers/PoisTest.java
+++ b/tiles/src/test/java/com/protomaps/basemap/layers/PoisTest.java
@@ -34,4 +34,48 @@ class PoisTest extends LayerTest {
         0
       )));
   }
+
+  @Test
+  void allotments() {
+    // this test shows two list elements because we're running against the whole profile, which means we're getting
+    // results form multiple layer classes. This may mean the test breaks when other layer classes are changed.
+    assertFeatures(15,
+      List.of(Map.of("pmap:kind", "allotments"),
+        Map.of("pmap:kind", "allotments", "pmap:min_zoom", 16, "name", "Kleingartenverein Kartoffel")),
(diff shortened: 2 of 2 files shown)


tiles/src/test/java/com/protomaps/basemap/layers/PoisTest.java
```diff
@@ -34,4 +34,43 @@ void busStop() {
...
+        "osm",
+        null,
+        0
+      )));
+  }
+
+  @Test
+  void playground() {
+    assertFeatures(15,
+      List.of(Map.of("pmap:kind", "playground"), Map.of("pmap:kind", "playground", "pmap:min_zoom", 13, "name", "Spielwiese")),
```
lenalebt, line 67: I am a bit unsure about why I see two elements here - maybe just have a closer look whetehr the tests are what you would expect :-)

lenalebt:
Will fix the issues monday or tuesday, won't find time beforehand

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -355,7 +356,9 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
             }
           }
         } else if (kind.equals("cemetery") ||
-          kind.equals("school")) {
+          kind.equals("school") ||
+          kind.equals("allotments") ||
```
nvkelso, line 360: Allotment POIs should appear later than schools, more like zoom 15 or 16

(see also how Tilezen sets the value: https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1258-L1266)

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -355,7 +356,9 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
...
+          kind.equals("playground")) {
```
nvkelso, line 361: Playground POIs should only appear at later zooms, like zoom 16.

(see also how Tilezen sets the value: https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1315-L1324)

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -323,7 +323,8 @@ public void processFeature(SourceFeature sf, FeatureCollector features) {
         } else if (kind.equals("forest") ||
           kind.equals("park") ||
           kind.equals("protected_area") ||
-          kind.equals("nature_reserve")) {
+          kind.equals("nature_reserve") ||
+          kind.equals("village_green")) {
```
nvkelso, line 327: This could work out... but might need some adjustments. Generally village greens are pretty small, and the other features in this grouping can get much larger.

Most similar Tilezen section is here https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L143-L152, but there's some related clamping for `commons`.

tiles/src/test/java/com/protomaps/basemap/layers/PoisTest.java
nvkelso: @bdon can you comment on this question about test construction, please?

bdon: The tests run against the entire Profile, not just a single layer class, so you're also getting the landuse element in the result set.

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
lenalebt: Then I'll add it such that allotments larger than 250m² get shown from zoom level 15, and smaller ones from 16 (basically last part of this branch).

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
lenalebt: Okay, I'll add a separate branch showing them from zoom level 16 onwards.

tiles/src/test/java/com/protomaps/basemap/layers/PoisTest.java
lenalebt: Okay, understood, but that means I'm basically not unit-testing the part I'm interested in. I don't necessarily want to refactor it so it is independent - I think it's rather up to you how you want this part to evolve in the future. I'll add a comment about what is happening.

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
lenalebt: What adjustments are you thinking about?

bdon:
@lenalebt thanks for adding these - can you suggest a representative area of the world so I can try out the PR on it? 

The nature of tileset changes is that they will affect the visual output. One of the things I'm working on is a test suite that lets us regression compare visually any two states in the past. Example of a test for bridges: https://maps.protomaps.com/visualtests/?tag=bridge

Ideally a visual test example for this would have one or more playgrounds and village greens / allotments

bdon:
(Nevermind, I see your link for https://www.openstreetmap.org/#map=18/51.50497/7.31888 , perfect example!)

bdon:
Merging for now, will follow up with style changes later, thanks!

Issue #224, opened 2024-03-26
More baselayers


pietervdvn:
Hi,

Because I didn't like the colour scheme of 'protomaps.light', I'm in the process of creating a new one - partly inspired by Mapnik/Osm-carto, partly by MapTiler. You can already browse it at https://dev.mapcomplete.org/cyclofix.html? (ignore the pins)

It uses the current hosted version of protomaps.

Would this be considered a nice addition for protomaps? If so, please point me where this theme should be put, I'll open a PR then.

bdon:
In general, no, themes beyond the base 5 (light, dark, white, grayscale, black) we can't commit to maintaining indefinitely into the future. 

The suggested theming approach is to create a set of named color variables like here https://github.com/protomaps/basemaps/blob/main/styles/src/themes.ts instead of writing a complete style, because that would let any color scheme take advantage of changes in the upstream style.

Otherwise you can suggest an improvement specific to what you don't like? Deriving things from OSM Carto is fine, but other styles will likely not be license-compatible with the styles in this repo.

Edefritz:
Thanks for your suggestions about creating a theme instead of meddling with the styles themselves.

But as far as I can see, there is no way to pass a custom theme to any of the exported functions of the package. They basically only allow passing a key to existing themes:

```ts
export default function (source: string, key: string): LayerSpecification[] {
  const theme = themes[key];
  return nolabels_layers(source, theme).concat(labels_layers(source, theme));
}

export function noLabels(source: string, key: string): LayerSpecification[] {
  const theme = themes[key];
  return nolabels_layers(source, theme);
}

export function labels(source: string, key: string): LayerSpecification[] {
  const theme = themes[key];
  return labels_layers(source, theme);
}

```

I guess to do that we need something like this?

```ts
export function layersWithTheme(source: string, theme: Theme): LayerSpecification[] {
  return nolabels_layers(source, theme).concat(labels_layers(source, theme));
}
```

I'm happy to provide a PR if necessary. Or am I missing something?

bdon:
@Edefritz yes, that looks like the right approach. It doesn't express a good way to separate labels/no labels for custom themes, or a way to inherit parts of a theme without re-defining every single theme property, but we can address those issues later. Can you open a PR?

bdon:
I've gotten a lot of feedback wanting more custom styles, so I've created a repository for "3rd-party" styles here: 

https://github.com/protomaps/basemaps-themes

[...]

PR #234, opened 2024-04-08
Allow passing custom themes to create layers

Edefritz:
So far it was only possible to generate styles based on a set of fixed themes maintained by the protomaps project.

With this change, it is possible to pass a custom colour theme to create layers, without interfering with the final style. Also passing partial themes is possible to override only a small set of properties.

Example:

```ts
const myCustomTheme: Theme = {...}
const style = layersWithCustomTheme("protomaps", myCustomTheme)
```
or
```ts
const partialTheme: Partial<Theme> = { background: "#fff" };
const style = layersWithPartialCustomTheme(
    "protomaps",
    "dark", // use dark theme as base
    partialTheme
);
```

Related issue: https://github.com/protomaps/basemaps/issues/224

This is just my take on addressing this problem. Please let me know if you have any suggestions or if you think some of these changes aren't necessary. And feel free to edit or rename anything.

Thanks :)

### DIFF
--- a/styles/src/index.ts
+++ b/styles/src/index.ts
@@ -17,19 +17,34 @@ export function labels(source: string, key: string): LayerSpecification[] {
   return labels_layers(source, theme);
 }
 
-export function layersWithCustomTheme(source: string, theme: Theme): LayerSpecification[] {
+export function layersWithCustomTheme(
+  source: string,
+  theme: Theme,
+): LayerSpecification[] {
   return nolabels_layers(source, theme).concat(labels_layers(source, theme));
 }
 
-export function layersWithPartialCustomTheme(source: string, key: string, partialTheme: Partial<Theme>): LayerSpecification[] {
+export function layersWithPartialCustomTheme(
+  source: string,
+  key: string,
+  partialTheme: Partial<Theme>,
+): LayerSpecification[] {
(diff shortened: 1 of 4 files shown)


bdon:
I like the partial theme solution, though I'm trying to imagine the use case where you want to change only one color - it seems more likely we will want a way to selectively remove a layer completely instead of just change the color values. Is the partial theme solution as implemented here enough for your use case?

styles/src/index.ts
```diff
@@ -16,3 +16,35 @@ export function labels(source: string, key: string): LayerSpecification[] {
...
+  key: string,
+  partialTheme: Partial<Theme>,
+): LayerSpecification[] {
+  const mergedTheme = { ...themes[key], ...partialTheme };
+  return nolabels_layers(source, mergedTheme).concat(
+    labels_layers(source, mergedTheme),
+  );
+}
+
+export function noLabelsWithCustomTheme(
```
bdon, line 38: I would prefer not to have these two functions that just call another function with the same arguments.

bdon: I'm going to merge this for now, we can think about a cleaner API in styles 4.0 once we understand the use cases better :P

bdon:
Thanks!

Issue #231, opened 2024-04-05
TS and JS build files of npm package out of sync


Edefritz:
Hi,

I discovered something unusual. When installing the latest protomaps-themes-base@2.0.0-alpha.5, I expected that the improvement [I suggested a while ago](https://github.com/protomaps/basemaps/pull/192) would be reflected.

According to the base_layers.ts [it is](https://github.com/protomaps/basemaps/blob/main/styles/src/base_layers.ts#L2002). But checking the style after I updated the package in my code base, the output seemed exactly the same as before.

After a lot of back and forth debugging, I found that the compiled .js, .mjs and .cjs files in the node_modules folder, didn't seem to have been updated. 

To rule out issues on my local installation, I checked the source of the package inside NPM: https://www.npmjs.com/package/protomaps-themes-base?activeTab=code

As you can see in the screenshots the contents of the JS and TS files don't match (zoom should be 6 in line 1927).

<img width="586" alt="Screenshot 2024-04-05 at 13 57 36" src="https://github.com/protomaps/basemaps/assets/1841670/8af45940-08d1-454d-8a9a-72a8419dd2fb">
<img width="607" alt="Screenshot 2024-04-05 at 13 57 10" src="https://github.com/protomaps/basemaps/assets/1841670/68d9dc4a-6c75-4719-8927-ffdb221fe26d">

I'm not sure this is how it works but perhaps you didn't compile the files before publishing the package? Or I'm awfully lost 😄

bdon:
Can you try 2.0.0-alpha.6?

I made some fixes to package.json, it might have been picking up an older CJS file by default

https://unpkg.com/browse/protomaps-themes-base@2.0.0-alpha.6/dist/index.js linę 1941 should be fixed
publint reports no issues: https://publint.dev/protomaps-themes-base@2.0.0-alpha.6

Edefritz:
Thanks so much for the quick response. But somehow this gives me different issues. If I just swap protomaps-themes-base@2.0.0-alpha.5 with protomaps-themes-base@2.0.0-alpha.6 I get the following error:

```js
$ bun run src/index.ts
1 | (function (entry, fetcher)
    ^
SyntaxError: Missing 'default' export in module '{...}/node_modules/protomaps-themes-base/dist/index.js'.
```

My project runs in bun, but I also tested it with an empty Node project. 

[...]

PR #248, opened 2024-05-14
Add CommonJS build target back. [#231]

bdon:
* emit the CommonJS build as index.cjs.

### DIFF
--- a/styles/package-lock.json
+++ b/styles/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "protomaps-themes-base",
-  "version": "3.0.0",
+  "version": "3.0.1",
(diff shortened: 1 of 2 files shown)


bdon:
Current issue:

```
npx publint
protomaps-themes-base lint results:
Warnings:
1. pkg.exports.require types is not exported. Consider adding pkg.exports.require.types to be compatible with TypeScript's "moduleResolution": "bundler" compiler option. Note that you cannot use "./dist/index.d.ts" because it has a mismatching format. Instead, you can duplicate the file and use the .cts extension, e.g. pkg.exports.require.types: "./dist/index.d.cts"
```

Issue #253, opened 2024-06-01
Add script information of `name` tag


wipfli:
**Description**
I would like to change the tiles to include information about the script/writing system used in the `name` tag. This will be useful to have language localized styles.

PR #254, opened 2024-06-01
Add pmap:script

wipfli:
Implements #253

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -142,7 +142,7 @@ public class Places implements ForwardingProfile.FeatureProcessor, ForwardingPro
       if (!script.equals("LATIN") && !script.equals("GENERIC")) {
         feat.setAttr("pmap:script", script);
       }
-      
+
       // NOTE: The buffer needs to be consistent with the innteral grid pixel sizes
       feat.setPointLabelGridPixelSize(LOCALITY_GRID_SIZE_ZOOM_FUNCTION)
         .setPointLabelGridLimit(LOCALITY_GRID_LIMIT_ZOOM_FUNCTION)
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -139,7 +139,7 @@ public class Places implements ForwardingProfile.FeatureProcessor, ForwardingPro
         .setSortKey(getSortKey(minZoom, 2, populationRank, population, sf.getString("name")));
 
       var script = Script.getScript(sf.getTag("name").toString());
-      if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+      if (!script.equals("Latin") && !script.equals("Generic")) {
         feat.setAttr("pmap:script", script);
       }
 
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -12,6 +12,7 @@ import com.protomaps.basemap.feature.FeatureId;
 import com.protomaps.basemap.feature.NaturalEarthDb;
 import com.protomaps.basemap.names.NeNames;
 import com.protomaps.basemap.names.OsmNames;
+import com.protomaps.basemap.names.Script;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -137,6 +138,11 @@ public class Places implements ForwardingProfile.FeatureProcessor, ForwardingPro
         // since all these are locality, we hard code kindRank to 2 (needs to match OSM section below)
         .setSortKey(getSortKey(minZoom, 2, populationRank, population, sf.getString("name")));
 
+      var script = Script.getScript(sf.getTag("name").toString());
+      if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+        feat.setAttr("pmap:script", script);
+      }
+      
       // NOTE: The buffer needs to be consistent with the innteral grid pixel sizes
       feat.setPointLabelGridPixelSize(LOCALITY_GRID_SIZE_ZOOM_FUNCTION)
         .setPointLabelGridLimit(LOCALITY_GRID_LIMIT_ZOOM_FUNCTION)
--- a/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
@@ -4,8 +4,6 @@ import com.onthegomap.planetiler.FeatureCollector;
 import com.onthegomap.planetiler.reader.SourceFeature;
 import java.util.Map;
 
-import com.protomaps.basemap.names.Script;
-
 public class NeNames {
 
   private NeNames() {}
--- a/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
@@ -4,6 +4,8 @@ import com.onthegomap.planetiler.FeatureCollector;
 import com.onthegomap.planetiler.reader.SourceFeature;
 import java.util.Map;
 
+import com.protomaps.basemap.names.Script;
+
 public class NeNames {
 
   private NeNames() {}
@@ -14,6 +16,11 @@ public class NeNames {
       var key = tag.getKey();
       if (key.equals("name")) {
         feature.setAttrWithMinzoom(key, sf.getTag(key), minZoom);
+
+        var script = Script.getScript(sf.getTag(key).toString());
+        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+          feature.setAttrWithMinzoom("pmap:script", script, minZoom);
+        }
       } else if (key.startsWith("name_")) {
         feature.setAttrWithMinzoom(key.replace("_", ":"), sf.getTag(key), minZoom);
       }
--- a/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
@@ -16,7 +16,7 @@ public class NeNames {
         feature.setAttrWithMinzoom(key, sf.getTag(key), minZoom);
 
         var script = Script.getScript(sf.getTag(key).toString());
-        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+        if (!script.equals("Latin") && !script.equals("Generic")) {
           feature.setAttrWithMinzoom("pmap:script", script, minZoom);
         }
       } else if (key.startsWith("name_")) {
--- a/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
@@ -6,8 +6,6 @@ import java.text.NumberFormat;
 import java.text.ParsePosition;
 import java.util.Map;
 
-import com.protomaps.basemap.names.Script;
-
 public class OsmNames {
 
   private OsmNames() {}
--- a/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
@@ -21,7 +21,7 @@ public class OsmNames {
 
       if (key.equals("name")) {
         var script = Script.getScript(sf.getTag(key).toString());
-        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+        if (!script.equals("Latin") && !script.equals("Generic")) {
           feature.setAttrWithMinzoom("pmap:script", script, minZoom);
         }
       }
--- a/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
+++ b/tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
@@ -6,6 +6,8 @@ import java.text.NumberFormat;
 import java.text.ParsePosition;
 import java.util.Map;
 
+import com.protomaps.basemap.names.Script;
+
 public class OsmNames {
 
   private OsmNames() {}
@@ -17,6 +19,13 @@ public class OsmNames {
       // Full names of places (default and translations)
       if (key.equals("name") || key.startsWith("name:")) {
         feature.setAttrWithMinzoom(key, sf.getTag(key), minZoom);
+      }
+
+      if (key.equals("name")) {
+        var script = Script.getScript(sf.getTag(key).toString());
+        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+          feature.setAttrWithMinzoom("pmap:script", script, minZoom);
+        }
       }
     }
     return feature;
--- a/tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
+++ b/tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
@@ -1,0 +1,85 @@
+package com.protomaps.basemap.names;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+class ScriptTest {
+
+  @Test
+  void singleScript() {
+    assertEquals("LATIN", Script.getScript("Berlin"));
+    assertEquals("ARABIC", Script.getScript("دبي"));
+    assertEquals("CYRILLIC", Script.getScript("Скопје"));
+    assertEquals("HAN", Script.getScript("北京"));
+    assertEquals("ETHIOPIC", Script.getScript("አዲስ አበባ"));
+    assertEquals("ARMENIAN", Script.getScript("Երևան"));
+    assertEquals("HANGUL", Script.getScript("서울"));
+    assertEquals("GEORGIAN", Script.getScript("თბილისი"));
+    assertEquals("GREEK", Script.getScript("Αθήνα"));
+    assertEquals("MONGOLIAN", Script.getScript("ᠤᠯᠠᠭᠠᠨᠪᠠᠭᠠᠲᠤᠷ"));
+    assertEquals("DEVANAGARI", Script.getScript("काठमाडौँ"));
+    assertEquals("HEBREW", Script.getScript("תל אביב"));
+    assertEquals("KANNADA", Script.getScript("ಬೆಂಗಳೂರು"));
+    assertEquals("BENGALI", Script.getScript("ঢাকা"));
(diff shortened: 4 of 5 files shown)


wipfli:
I generated a Europe pmtiles file with this change and made a style.json localized to English.

### Changes in the style.json

`"text-field": "{name:en}"` in the country labels layer

And in the places_locality layer I used this:

```json
"text-field": [
  "format",
  [
      "case",
      ["!", ["has", "pmap:script"]],
      ["coalesce", ["get", "name:en"], ["get", "name:de"], ["get", "name"], ""],
      ["has", "pmap:script"],
      ["coalesce", ["get", "name:en"], ["get", "name:de"], ""],
      ""
  ],
  {},
  "\n",
  {},
  [
      "case",
      ["has", "pmap:script"],
      ["get", "name"],
      ""
  ],
  {}
]
```

nvkelso:
Nice, the demo looks great!

wipfli:
### Style localized to Greek

`"text-field": "{name:el}"` in the country labels layer

And in the places_locality layer I used this:

```json
"text-field": [
  "format",
  [
      "case",
      ["==", ["get", "pmap:script"], "GREEK"],
      ["coalesce", ["get", "name:el"], ["get", "name"], ""],
      ["!=", ["get", "pmap:script"], "GREEK"],
      ["coalesce", ["get", "name:el"], ""],
      ""
  ],
  {},
  "\n",
  {},
  [
      "case",
      ["!=", ["get", "pmap:script"], "GREEK"],
      ["get", "name"],
      ""
  ],
  {}
]
```
#### Athens

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/el.html#map=8.55/37.9815/23.801

image


#### Skopie

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/el.html#map=10.68/41.9835/21.5161

image

#### Berlin

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/el.html#map=9.54/52.5462/13.4908

image

wipfli:
### Style localized to Bulgarian

`"text-field": "{name:bg}"` in the country labels layer

And in the places_locality layer I used this:

```json
"text-field": [
  "format",
  [
      "case",
      ["==", ["get", "pmap:script"], "CYRILLIC"],
      ["coalesce", ["get", "name:bg"], ["get", "name:ru"], ["get", "name:uk"], ["get", "name:sr"], ["get", "name:mk"], ["get", "name:be"], ["get", "name:kk"], ["get", "name:ky"], ["get", "name"], ""],
      ["!=", ["get", "pmap:script"], "CYRILLIC"],
      ["coalesce", ["get", "name:bg"], ["get", "name:ru"], ["get", "name:uk"], ["get", "name:sr"], ["get", "name:mk"], ["get", "name:be"], ["get", "name:kk"], ["get", "name:ky"], ""],
      ""
  ],
  {},
  "\n",
  {},
  [
      "case",
      ["!=", ["get", "pmap:script"], "CYRILLIC"],
      ["get", "name"],
      ""
  ],
  {}
],
```

#### Athens

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/bg.html#map=8.55/37.9815/23.801

image

#### Skopie

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/bg.html#map=10.68/41.9835/21.5161

image

#### Berlin

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/bg.html#map=9.54/52.5462/13.4908

image

bdon:
I wonder if this:

`["!=", ["get", "pmap:script"], "CYRILLIC"],`

Is there some way to move the logic of what language uses what script into the tiles? such as:

`["!=", ["get", "pmap:script"], ["get","pmap:...."]],`

Otherwise we'd need to encode a map of `bg` -> `CYRILLIC` in the frontend code, right?

Is there a standardized script code instead of `LATIN`, `CYRILLIC` somewhere in Harfbuzz?

wipfli:
> Is there a standardized script code instead of LATIN, CYRILLIC somewhere in Harfbuzz?

We could use the ISO15924 script codes consisting of 4 letters, see https://unicode.org/iso15924/iso15924-codes.html. Harfbuzz uses those as well. Note however that the java function `Character.UnicodeScript.of` already uses the standardized English names of the scripts from unicode, see https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/lang/Character.UnicodeScript.html

> Is there some way to move the logic of what language uses what script into the tiles?

I think it makes most sense to decide in the style json what script can be used for a given language. The script is a characteristic of the language and not of the individual feature in a vector tile...

Below you find an attempt to group languages by script. Note that some languages can be written in more than one script, e.g., Malay can be written in Latin, Arabic, and Thai.


### Scripts Supported by MapLibre


#### Latin



* AFRIKAANS
* ALBANIAN
* AZERBAIJANI - also cyrillic, arabic
* BASQUE
* BOSNIAN - also cyrillic
* CATALAN
* CROATIAN
* CZECH
* DANISH
* DUTCH
* ENGLISH
* ENGLISH (AUSTRALIAN)
* ENGLISH (GREAT BRITAIN)
* ESTONIAN
* FINNISH
* FILIPINO
* FRENCH
* FRENCH (CANADA)
* GALICIAN
* GERMAN
* HUNGARIAN
* ICELANDIC
* INDONESIAN
* ITALIAN
* LATVIAN
* LITHUANIAN
* MALAY - also arabic, thai
* NORWEGIAN
* POLISH
* PORTUGUESE
* PORTUGUESE (BRAZIL)
* PORTUGUESE (PORTUGAL)
* ROMANIAN
* SLOVAK - also cyrillic
* SLOVENIAN
* SPANISH
* SPANISH (LATIN AMERICA)
* SWAHILI
* SWEDISH
* TURKISH
* UZBEK - also cyrillic, arabic 
* VIETNAMESE
* ZULU


#### Arabic



* ARABIC
* FARSI
* URDU


#### Cyrillic



* BELARUSIAN
* BULGARIAN - also latin
* KAZAKH - also latin, arabic
* KYRGYZ - also arabic
* MACEDONIAN
* RUSSIAN
* SERBIAN - also latin
* UKRAINIAN


#### Han ?



* CHINESE
* CHINESE (SIMPLIFIED)
* CHINESE (HONG KONG)
* CHINESE (TRADITIONAL)


#### One Language Per Script



* AMHARIC
* ARMENIAN
* KOREAN
* JAPANESE
* GEORGIAN
* GREEK
* MONGOLIAN - also cyrillic


### Scripts Not Supported by MapLibre


#### Devanagari



* GUJARATI
* HINDI
* MARATHI
* NEPALI


#### One Language Per Script



* HEBREW
* KANNADA
* BENGALI
* BURMESE
* KHMER
* LAO
* MALAYALAM
* PUNJABI
* SINHALESE
* TAMIL
* TELUGU
* THAI

wipfli:
I did some experimentation with the normal protomaps tiles and the normal protomaps style, but added a javascript hook to MapLibre GL JS which allows you to modify vector tile feature properties in the client.

https://wipfli.github.io/maplibre-feature-properties-transform-example/language?script=Latin&language=fr#map=8/38/24

You can choose different languages. The logic is contained [here](https://github.com/wipfli/maplibre-feature-properties-transform-example/blob/main/language-transform.js) and uses some regular expressions on the `name` tag.

The main problem I see at the moment is when the `name` tag contains text in multiple scripts. For example, [OSM Relation 11118019](https://www.openstreetmap.org/relation/11118019) has 

```
name = Ngushtica e Otrantos / Canale d'Otranto / Στενό του Οτράντο
```

So it contains text in Latin and Greek.

wipfli:
... and I delete text from the `name` name tag which has the same script as the chosen language. For example, when the language is set to German, the result is this:

image

https://wipfli.github.io/maplibre-feature-properties-transform-example/language?script=Latin&language=de#map=8/39.996/19.233

nvkelso:
This isn't working for me in safari or chrome on iOS.

wipfli:
Looks like also firefox does not like the syntax `data:,<your-js-code>` for dataURL containing javascript. I replaced it now with
```js
const decodedString = decodeURIComponent(`${settings}${transform}`);
const blob = new Blob([decodedString], { type: 'application/javascript' });
const workerUrl = URL.createObjectURL(blob);
maplibregl.importScriptInWorkers(workerUrl);
```

wipfli:
Sorry for getting side-tracked by the MapLibre GL JS feature properties transform thing. I think for now it would be best to move on with this pull request as is.

1. Do you want me to shorten things like `LATIN` to `LATN` as defined [here](https://unicode.org/iso15924/iso15924-codes.html)? I am against it because tile size savings are minimal due to string indexing in MVT and the readability gets worse.
2. Do I need to add some tests?

wipfli:
I am a bit lost in spotless:apply.

wipfli:
@bdon this is ready for review from my side.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertEquals("LAO", Script.getScript("ປາກເຊ"));
+    assertEquals("MALAYALAM", Script.getScript("കൊല്ലം"));
+    assertEquals("GURMUKHI", Script.getScript("ਜਲੰਧਰ"));
+    assertEquals("SINHALA", Script.getScript("කොළඹ"));
+    assertEquals("TAMIL", Script.getScript("கொழும்பு"));
+    assertEquals("TELUGU", Script.getScript("హైదరాబాద్"));
+    assertEquals("THAI", Script.getScript("กรุงเทพมหานคร"));
+  }
+
+  @Test
```
bdon, line 36: I like these test cases, what about mixed Han scripts like 台北101 (should be HAN) or any of these:

https://en.wikipedia.org/wiki/Hiragana_and_katakana_place_names

I am still thinking if there is a way for us to do the fallback-ing on the tile side instead of making the client style more complicated. but that might require `name:latin`, name:cyrillic` (mentioned this to @1ec5) - anyways I believe adding the `pmap:script` for non-latin is necessary for now to make progress (we can trim things down in an upcoming major version).

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -137,6 +138,11 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
         // since all these are locality, we hard code kindRank to 2 (needs to match OSM section below)
         .setSortKey(getSortKey(minZoom, 2, populationRank, population, sf.getString("name")));
 
+      var script = Script.getScript(sf.getTag("name").toString());
+      if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+        feat.setAttr("pmap:script", script);
```
nvkelso, line 143: Generally we prefer lowercase keys and values elsewhere in the tile schema.

But if the [ISO standard](https://unicode.org/iso15924/iso15924-codes.html) prefers Titlecases then that's OK. But I don't see UPPERCASE in the standard.

tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
```diff
@@ -14,6 +14,11 @@ public static FeatureCollector.Feature setNeNames(FeatureCollector.Feature featu
       var key = tag.getKey();
       if (key.equals("name")) {
         feature.setAttrWithMinzoom(key, sf.getTag(key), minZoom);
+
+        var script = Script.getScript(sf.getTag(key).toString());
+        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
```
nvkelso, line 19: See above re value casing.

tiles/src/main/java/com/protomaps/basemap/names/NeNames.java
```diff
@@ -14,6 +14,11 @@ public static FeatureCollector.Feature setNeNames(FeatureCollector.Feature featu
...
+          feature.setAttrWithMinzoom("pmap:script", script, minZoom);
```
nvkelso, line 20: See above re value casing.

tiles/src/main/java/com/protomaps/basemap/names/OsmNames.java
```diff
@@ -18,6 +18,13 @@ public static FeatureCollector.Feature setOsmNames(FeatureCollector.Feature feat
       if (key.equals("name") || key.startsWith("name:")) {
         feature.setAttrWithMinzoom(key, sf.getTag(key), minZoom);
       }
+
+      if (key.equals("name")) {
+        var script = Script.getScript(sf.getTag(key).toString());
+        if (!script.equals("LATIN") && !script.equals("GENERIC")) {
+          feature.setAttrWithMinzoom("pmap:script", script, minZoom);
```
nvkelso, line 25: See above re value casing.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.jupiter.api.Test;
+
+class ScriptTest {
+
+  @Test
+  void singleScript() {
+    assertEquals("LATIN", Script.getScript("Berlin"));
+    assertEquals("ARABIC", Script.getScript("دبي"));
+    assertEquals("CYRILLIC", Script.getScript("Скопје"));
+    assertEquals("HAN", Script.getScript("北京"));
```
nvkelso, line 14: We should try to be explicate about Simplified Chinese versus Traditional Chinese

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
nvkelso: This Java library returns UPPERCASE constants for the various scripts. So I lean towards lowercasing them in our export. Something like:

> var script = Script.getScript(sf.getTag("name").toString().toLowerCase());

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
nvkelso: And add a comment this is {Simplified|Traditional} "Chinese, please.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class ScriptTest {
+
+  @Test
+  void singleScript() {
+    assertEquals("LATIN", Script.getScript("Berlin"));
+    assertEquals("ARABIC", Script.getScript("دبي"));
+    assertEquals("CYRILLIC", Script.getScript("Скопје"));
```
nvkelso, line 13: Might add a comment this is for Russian and related languages?

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertEquals("MYANMAR", Script.getScript("ရန်ကုန်"));
+    assertEquals("KHMER", Script.getScript("ភ្នំពេញ"));
+    assertEquals("LAO", Script.getScript("ປາກເຊ"));
+    assertEquals("MALAYALAM", Script.getScript("കൊല്ലം"));
+    assertEquals("GURMUKHI", Script.getScript("ਜਲੰਧਰ"));
+    assertEquals("SINHALA", Script.getScript("කොළඹ"));
+    assertEquals("TAMIL", Script.getScript("கொழும்பு"));
+    assertEquals("TELUGU", Script.getScript("హైదరాబాద్"));
+    assertEquals("THAI", Script.getScript("กรุงเทพมหานคร"));
+  }
```
nvkelso, line 34: Might add entry for Japanese?

There are 4, sorry:

Japanese syllabaries (alias for Han + Hiragana + Katakana)

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  void latinPunctuation() {
+    assertEquals("LATIN", Script.getScript("Berlin!"));
+  }
+
+  @Test
+  void nonAsciiLatin() {
+    assertEquals("LATIN", Script.getScript("Zürich"));
```
nvkelso, line 73: Great test, thanks!

wipfli: :+1:

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
wipfli: Good point. I checked on Wikipedia and there they use always first letter uppercase followed by lowercase letters, e.g., "Devanagari". This makes sense to me since languages are similarly capitalized. I adopted the code.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
wipfli: Added those.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
wipfli: Added a test for `台北101` - it is just Han.

Regarding `name:latin`, I agree that this could be an interesting addition. But it solves a different problem than this pull request. Here the goal is really to know if the content of the `name` tag is in a certain script whereas with `name:latin` one could shorten the language fallback chain significantly.

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
wipfli: I think unicode actually uses the same codepoints for those, see https://en.wikipedia.org/wiki/Han_unification. I could not find any counter examples but I am happy to be proven wrong. Added a code comment for now...

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
wipfli: Added

tiles/src/test/java/com/protomaps/basemap/names/ScriptTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  void singleScript() {
+    assertEquals("LATIN", Script.getScript("Berlin"));
+    assertEquals("ARABIC", Script.getScript("دبي"));
+    assertEquals("CYRILLIC", Script.getScript("Скопје"));
+    assertEquals("HAN", Script.getScript("北京"));
+    assertEquals("ETHIOPIC", Script.getScript("አዲስ አበባ"));
+    assertEquals("ARMENIAN", Script.getScript("Երևան"));
+    assertEquals("HANGUL", Script.getScript("서울"));
```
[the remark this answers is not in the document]
wipfli: Added

wipfli:
Thanks for the reviews! I think I addressed all comments. How should I handle the failing sonarcloud CI?

PR #265, opened 2024-06-23
Use a positioned glyph font encoding for Devanagari

wipfli:
Encodes parts of strings that are in the Devanagari script with a positioned glyph font from https://github.com/wipfli/pgf-encoding.

### Engish
To show a map localized to English one can do then something like this:

```jsonc
"text-field": [
    "case",
    ["has", "pmap:script"],
    [
        "format",
        [
        "coalesce",
        ["get", "name:en"],
        ["get", "name:de"],
        ["get", "name:fr"]
        ],
        {},
        "\n",
        {},
        ["get", "pmap:pgf:name"],
        {
        "text-font": [
            "case",
            ["==", ["get", "pmap:script"], "Devanagari"],
            ["literal", ["NotoSansDevanagari-Regular-v1"]],
            ["literal", ["Noto Sans Regular"]]
        ]
        }
    ],
    [
        "coalesce",
        ["get", "name:en"],
        ["get", "name"]
    ]
],
```

### Nepali
For Nepali one can on the other hand do something like this:

```jsonc
"text-field": [
    "case",
    ["!=", ["get", "pmap:script"], "Devanagari"],
    [
        "format",
        [
        "coalesce",
        ["get", "pmap:pgf:name:ne"],
        ["get", "pmap:pgf:name:mr"],
        ["get", "pmap:pgf:name:hi"]
        ],
        {
        "text-font": ["literal", ["NotoSansDevanagari-Regular-v1"]]
        },
        "\n",
        {},
        ["get", "pmap:pgf:name"],
        {
        "text-font": [
            "case",
            ["==", ["get", "pmap:script"], "Devanagari"],
            ["literal", ["NotoSansDevanagari-Regular-v1"]],
            ["literal", ["Noto Sans Regular"]]
        ]
        }
    ],
    [
        "format",
        [
        "coalesce",
        ["get", "pmap:pgf:name:ne"],
        ["get", "pmap:pgf:name"]
        ],
        {
        "text-font": ["literal", ["NotoSansDevanagari-Regular-v1"]]
        }
    ]
],
```

### Planet PMTiles

https://pub-726b01260c98468a9387cc0dfcb7386b.r2.dev/protomaps-planet-pgf-20240627.pmtiles (126 GB)

### Nepali

Note: Only the `places_locality` layer was changed.

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/ne.html

### English

Note: Only the `places_locality` layer was changed.

https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/en.html

<a href="https://pub-cf7f11e26ace447db8f7215b61ac0eae.r2.dev/en.html#map=11.59/27.6914/85.3365">

</a>

## Todos

- [x] Remove all todo comments
- [x] Handle errors and exceptions
- [x] Add unit tests
- [x] Remove `main()` functions
- [x] Store fonts and versions in tilejson metadata


tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -138,11 +143,18 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
...
+      var script = Script.getScript(name);
+
       if (!script.equals("Latin") && !script.equals("Generic")) {
         feat.setAttr("pmap:script", script);
+        if (fontRegistry != null && fontRegistry.getScripts().contains(script)) {
+          name = TextEngine.encodeRegisteredScripts(name, fontRegistry);
+        }
       }
 
+      feat.setAttr("name", name);
```
bdon, line 156: Do you think we should leave the proper Unicode in `name` and then have logic in the style called name:mm:pgf` with the non-Unicode encoding? That would make the tileset more consistent for non-MapLibre use cases

tiles/src/main/java/com/protomaps/basemap/Basemap.java
```diff
@@ -137,15 +139,21 @@ static void run(Arguments args) {
...
 
     var tmpDir = nePath.resolveSibling(nePath.getFileName() + "-unzipped");
     var naturalEarthDb = NaturalEarthDb.fromSqlite(nePath, tmpDir);
     //    var qrankDb = QrankDb.fromCsv(sourcesDir.resolve("qrank.csv.gz"));
     var qrankDb = QrankDb.empty();
 
-    planetiler.setProfile(new Basemap(naturalEarthDb, qrankDb)).setOutput(Path.of(area + ".pmtiles"))
+    String pgfEncodingRepoHash = "e9c03fb";
+    FontRegistry fontRegistry = new FontRegistry(pgfEncodingRepoHash);
+    fontRegistry.loadFontBundle("NotoSansDevanagari-Regular", "1", "Devanagari");
```
bdon, line 154: it feels like we should probably just make FontRegistry a Singleton referenced statically, since there is only ever one instance, right? 

Passing it as a dependency to each layer could increase the testability, but I don't see us ever needing to pass a mock FontRegistry into a layer for layer-specific behavior.

tiles/src/main/java/com/protomaps/basemap/text/TextEngine.java
```diff
@@ -0,0 +1,190 @@
...
+        encodedText += TextEngine.encode(segment, fontRegistry.getFont(script), fontRegistry.getEncoding(script));
+      }
+      else {
+        encodedText += segment;
+      }
+    }
+    return encodedText;
+  }
+
+  public static void main(String[] args) {
```
bdon, line 157: feels like we can move this `main` into a small unit or integration test instead of having multiple `mains` in the program?

tiles/src/main/java/com/protomaps/basemap/Basemap.java
```diff
@@ -137,15 +139,21 @@ static void run(Arguments args) {
       .addGeoPackageSource("landcover", sourcesDir.resolve("daylight-landcover.gpkg"),
         "https://r2-public.protomaps.com/datasets/daylight-landcover.gpkg");
 
-    Downloader.create(planetiler.config()).add("ne", neUrl, nePath).run();
+    Downloader.create(planetiler.config()).add("ne", neUrl, nePath)
+      // .add("pgf-encoding", "https://api.github.com/repos/wipfli/pgf-encoding/zipball/e9c03fb", sourcesDir.resolve("main.zip"))
```
bdon, line 143: Is the idea here the hash refers to a specific compilation of Unicode PUA -> precomposed glyph mappings, and if the hash changes, that means a major version bump and the encodings break? Is there an opportunity to SemVer those?

Is the hash version also associated with a set of Fontstacks (one for each supported script)?

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
wipfli: It is a good idea to keep the original Unicode string in `name` and `name:*`. I prefixed the encoded strings with `pmap:pgf:`.

Examples: 
* `name` contains original Unicode
* `pmap:pgf:name` contains parts that are in a registered script in the encoded string
* `name:ne` contains original Unicode
* `pmap:pgf:name:ne` contains parts that are in a registered script in the encoded string

tiles/src/main/java/com/protomaps/basemap/Basemap.java
wipfli: The static Singleton pattern seems like a good solution in this case. Thank you for the tip.

tiles/src/main/java/com/protomaps/basemap/text/TextEngine.java
wipfli: Moved to unit tests.

tiles/src/main/java/com/protomaps/basemap/Basemap.java
wipfli: I removed the commit has altogether and added some comments to the semver markdown file.

* The basemap major version has to bump when we change the version of a script that we already include
* The basemap minor version has to bump when we add a new script

wipfli:
Thank you for the great feedback @bdon. I think I address all concerns. The naming has been updated and I am currently running a planet build which should be ready for an updated demo in a few hours.

Let me know what else needs to change...

tiles/src/main/java/com/protomaps/basemap/Basemap.java
```diff
@@ -112,6 +116,21 @@ public String attribution() {
...
 
+  @Override
+  public Map<String, String> extraArchiveMetadata() {
+    Map<String, String> result = new HashMap<>();
+
+    FontRegistry fontRegistry = FontRegistry.getInstance();
+    List<String> scripts = fontRegistry.getScripts();
+
+    for (String script : scripts) {
+      result.put("pgf:" + script.toLowerCase() + ":name", fontRegistry.getName(script));
```
nvkelso, line 127: It wasn't clear to me that `pgf` stands for "positioned glyph font", consider adding a comment above this line to explain that?

wipfli: Good point. I was also wondering if we have a schema documentation somewhere?

nvkelso: In addition to the inline comment in the code here, we should probably add a new "names.md" or `localization.md` file next to https://github.com/protomaps/docs/blob/main/basemaps/layers.md in the Protomaps docs repo?

Tilezen had [this](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#feature-names), which we should only lightly crib from. 

Natural Earth's [table of common languages](https://github.com/nvkelso/natural-earth-vector/blob/master/packages/Natural_Earth_quick_start/LOCALIZATION.md#setting-names-localization) is useful, too, and add a script column to that?

image

...snip...

nvkelso: Stubbed out in new docs PR: https://github.com/protomaps/docs/pull/43/files

PR #267, opened 2024-06-27
Add better inline docs for langauage scripts

nvkelso:
Taken from this Github comment https://github.com/protomaps/basemaps/pull/254#issuecomment-2144007634 (which was getting lost) and additional research


nvkelso:
The other ones seem like single language scripts so choosing not to expand them with a redundant comment

wipfli (approved):

Issue #282, opened 2024-08-12
Tiles and styles 4.0.0


bdon:
**DESCRIPTION**

**Breaking changes:**

* drop `pmap:` prefixes from tag names to align with tilezen.
* Remove deprecated OSM tags in favor of `kind`
* remove `physical_point`, `natural` and `physical_line` layers to align with tilezen. features go into the `landuse` or `water`, `poi` or `earth` layers.

**Resolve other issues:**

* #256
* #112 

@nvkelso @wipfli

wipfli:
For v4 I would like to segment names by script #273 and remove the `pgf:` prefix (store pgf encoded strings by default, have bare unicode in other property).

bdon:
@nvkelso we also need to collapse `roads` > `medium_road` into `major_road` (with appropriate `kind_detail`)

nvkelso:
While Tilezen 1.x series doesn't have a separate layer for OSM `natural` tagged features, in v2 we said those should go into a `landcover` layer in https://github.com/tilezen/vector-datasource/issues/2025 and related https://github.com/protomaps/basemaps/issues/154.

nvkelso:
> remove physical_point, natural and physical_line layers to align with tilezen. features go into the landuse or water, poi or earth layers.

Are you ok with mixed geometry types in the MVT layers like Tilezen had?

nvkelso:
> Remove deprecated OSM tags in favor of kind

Need to ensure `kind_detail` captures the appropriate raw OSM values. I _think_ it does today, but it's the combo of `kind` and `kind_detail` that makes dropping the raw tags possible.

bdon:
> Are you ok with mixed geometry types in the MVT layers like Tilezen had?

We need to add geometry type filters to the style.

> While Tilezen 1.x series doesn't have a separate layer for OSM natural tagged features, in v2 we said those should go into a landcover layer in https://github.com/tilezen/vector-datasource/issues/2025 and related https://github.com/protomaps/basemaps/issues/154.

I am already separating out daylight landcover into a new `landcover` that is not part of Tilezen. 

Plan is to keep OSM landcover in `landuse` for Tangram preservation reasons until 5.0.

bdon:
breaking changes to tiles merged in https://github.com/protomaps/basemaps/commit/55d0df23a0091f810eafbbd5b156b8783f181ef4

will follow up with style changes

bdon:
Both tiles and styles have been merged now.

Issue #256, opened 2024-06-10
Label of New York is blinking


wipfli:
The label of New York City is blinking during zoom.

- zoom 1 to 2: visible
- zoom 2 to 3.65: hidden
- zoom 3.65 to 4.65: visible
- zoom 4.65 to 4.85: hidden
- zoom 4.85 and more: visible

wipfli:
Probably this is due to collisions between the `places_locality` layer and the `places_region` layer.

bdon:
I noticed this is happening because we have NE localities conflicting with OSM ones.

We should take geometries only from OSM and augment them with NE properties / minzoom information.

wipfli:
The Delhi city label also gets collided away by a state label:

```
map=8.77/28.5699/77.0001
```

![image](https://github.com/protomaps/basemaps/assets/53421382/b2d63564-f1b9-48e9-b3cb-47a5f8fc4426)

nvkelso:
I wonder about styling all the labels in the `places` data layer in a single style layer, sorted my the `pmap:min_zoom`. Then the collisions would be deterministic (if a low-zoom locality is present and labeled before the region, it should win in a collision in later zooms with region, not just other locality features). This would require more complicated nested styling logic, but worth prototyping?

wipfli:
Interesting idea I think. Another option could be to disable certain state and country labels (e g.: Delhi, Vatican).

bdon:
Given the situation in #286 I think tile ordering isn't good enough, so we should do this in the style:

> I wonder about styling all the labels in the places data layer in a single style layer, sorted my the pmap:min_zoom

wipfli:
Fixed by #436

PR #289, opened 2024-08-22
Places: don't use Natural Earth at low zooms [#256, #282]

bdon:
This makes the `places` data rely purely on OSM for geometries, removing the buggy duplication of places and making name labels consistent at all zoom levels. @wipfli @nvkelso 

Since we allowlist the OSM name tags this keeps the tile size reasonable at low zooms 0-7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Tiles v4.0.0-alpha.3
+------
+- Replace Natural Earth places at low zooms with OSM [#289]
+
 Tiles v4.0.0-alpha.2
 ------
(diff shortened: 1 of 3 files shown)


wipfli:
Do you have a preview map?

bdon:
@wipfli https://protomaps.dev/~bdon/places.pmtiles (omitted other layers using a filtered PBF)

nvkelso (changes requested):
~~snip~~

nvkelso (approved):
My bad, looks like we're already doing that! So no regression :)

PR #291, opened 2024-08-22
Fix highways

wipfli:
Due to https://github.com/protomaps/basemaps/pull/285 we currently do not have highways in the tiles below z12.

This pull requests fixes that, but we might want to add some sort of test to avoid this in the future.


wipfli:
The `highway` attribute is needed in post-process here:

https://github.com/protomaps/basemaps/blob/8e982d76bf5ab64dec67822fb2e7c280565528c7/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java#L188-L195

and gets removed here:

https://github.com/protomaps/basemaps/blob/8e982d76bf5ab64dec67822fb2e7c280565528c7/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java#L198

PR #292, opened 2024-08-25
Update NPM release

wipfli:


.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,147 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_ORG_TOKEN }}
+
+      - name: Publish NPM package (pre-release)
+        if: ${{ steps.prepare_release.outputs.release_type == 'prerelease' }}
+        run: |
+          npm publish --tag next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_ORG_TOKEN }}
+
+      # - name: Tag commit and push
```
bdon, line 89: can we remove the commented out parts?

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,147 @@
...
+          npm run check
+          npm run test
+          npm run build
+
+      - name: Publish NPM package (regular)
+        if: ${{ steps.prepare_release.outputs.release_type == 'regular' }}
+        run: |
+          npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_ORG_TOKEN }}
```
bdon, line 80: This just requires me to generate a granular access token and save it in GitHub actions env, right?

wipfli: Yes. In the MapLibre org we have a token for NPM stored in the GitHub org secrets. Then we share that github org secrets with individual GitHub repositories that need access to the NPM token. Alternatively, if you do not have other repos that publish to NPM you can also just store the token in this github repo as a repo-level secret...

wipfli: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-an-organization

.github/workflows/release-npm.yml
wipfli: OK

PR #298, opened 2024-09-17
Update MapLibre GL to v4.7.0

wipfli:


wipfli:
Gives you the latest collision boxes. They lag a bit since a recent pull request, but the computation of collisions is more correct now. https://github.com/maplibre/maplibre-gl-js/pull/4071

Issue #286, opened 2024-08-21
Barcelona label blinking


wipfli:
**Location**
[Include a link to a public URL with coordinates in the hash and zoom level.
](https://maps.protomaps.com/#map=9.8/41.4354/2.2428&theme=light&tiles=https://build.protomaps.com/20240821.pmtiles)

**Screenshots**
z9.87:
![image](https://github.com/user-attachments/assets/d409d825-93fc-47ae-8dc4-7c42870485d3)

z10.78:
![image](https://github.com/user-attachments/assets/b9bedd37-d742-4844-8a84-702e2f29fdca)

**Required information**
 - Renderer version [e.g. MapLibre GL JS 3.3.0]
 - Tileset version and source
 - Style JSON or NPM version

wipfli:
The style does not honor `population_rank`, and adding `"symbol-sort-key": ["*", -1, ["get", "pmap:population_rank"]]` to the `places_locality` layer's layout properties fixes the issue. However, better I think would be to sort the features in the tiles according to `pmap:population_rank` to avoid this issue.

wipfli:
We are using `setSortKey` already here:

https://github.com/protomaps/basemaps/blob/0eeafa51cb02671615300e3433374eef66f1417a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java#L391

The planetiler docs state:

>For symbols (text/icons) where clients try to avoid label collisions, features are placed in the order they appear in each layer, so features that appear earlier (lower sort key) will show up at lower zoom levels than feature that appear later (higher sort key) in a layer.

wipfli:
l'Hospitalet de Llobregat:

```
id | 17593125925341
-- | --
name:ru | Оспиталет-де-Льобрегат
capital | 8
name:el | Οσπιταλέτ ντε Λιοβρεγάτ
name | l'Hospitalet de Llobregat
pmap:kind_detail | city
pmap:min_zoom | 8
pmap:kind | locality
pmap:pgf:name | l'Hospitalet de Llobregat
wikidata | Q15470
pmap:population_rank | 10
population | 274455
```
collides away the Barcelona label:
```
id | 17592338408581
-- | --
capital | 4
name:uk | Барселона
pmap:kind_detail | city
pmap:pgf:name:hi | 
name:el | Βαρκελώνη
name:it | Barcellona
pmap:min_zoom | 4
name:en | Barcelona
name:es | Barcelona
name:zh | 巴塞罗那
[...]

PR #300, opened 2024-09-18
Pre-collide places layer point features and reduce density / overlaps

### LINKED ISSUE #116: Too many places visible at zoom 8, 9, and 10
While fixing https://github.com/protomaps/basemaps/issues/89 to add more state (`region`) labels, I inadvertently also added more `locality` labels. This was OK in the United States and Europe, but looks far too dense in India and China.

The grid size should probably be coarser at these mid-zooms (to disallow smaller rural localities from backfilling less populated areas of the map), and the number of allowed place labels should be slightly reduced (This could improve flickering in Paris at zoom 10).

```
private static final ZoomFunction<Number> LOCALITY_GRID_SIZE_ZOOM_FUNCTION =
    ZoomFunction.fromMaxZoomThresholds(Map.of(
      6, 32,
      7, 64
    ), 0);

  private static final ZoomFunction<Number> LOCALITY_GRID_LIMIT_ZOOM_FUNCTION =
    ZoomFunction.fromMaxZoomThresholds(Map.of(
      6, 8,
      7, 6,
      9, 4
    ), 0);
```

### LINKED ISSUE #287: Label density can be optimized at z8
In particular at z8 many labels in urban areas cannot be rendered due to collisions. Below is an example of New York, rendered with maplibre `map.showSymbolCollisions = true`. Maybe we can optimize tile size and bring fewer labels to the renderer which means fewer collision computations. Could make sense to have a symbol pruning step in Planetiler (cc @msbarry).

image

nvkelso:
Fixes #287 and fixes #116 and related to label flickering in https://github.com/protomaps/basemaps/issues/286 and https://github.com/protomaps/basemaps/issues/256.

Adjusts the collision grid size and sort functions.

Testing in on the `california` make target, we see dramatic reduction in features from zooms 5 to 13, with the most reduction in zooms 8, 9, and 10:

```
                 z0    z1    z2    z3    z4    z5    z6    z7    z8    z9   z10   z11   z12   z13   z14   z15   all
  places after    0   858  2.6k  2.6k  3.4k  6.6k  5.5k  7.6k  7.3k  7.4k  5.2k  5.3k  5.4k    3k    2k  1.6k  7.6k
  places before   0  1.8k  2.6k  2.9k  4.3k  8.9k  6.5k   10k   16k   10k  7.3k  5.9k  6.9k  4.1k  2.2k  1.7k   16k
  ```
  
The archive size and feature count remained constant.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -36,13 +36,15 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
       // (nvkelso 20230803) floats with significant single decimal precision
       //                    but results in "Too many possible values"
       // Order ASCENDING (smaller manually curated Natural Earth min_zoom win over larger values, across kinds)
-      .orderByInt((int) minZoom, 0, 15)
+      // minZoom is a float with 1 significant digit for manually curated places
+      .orderByInt((int)(minZoom * 10), 0, 150)
       // Order ASCENDING (smaller values win, countries then locality then neighbourhood, breaks ties for same minZoom)
-      .thenByInt(kindRank, 0, 6)
+      .thenByInt(kindRank, 0, 12)
       // Order DESCENDING (larger values win, San Francisco rank 11 wins over Oakland rank 10)
-      .thenByInt(populationRank, 15, 0)
+      // Disabled to allow population log to have larger range
+      //.thenByInt(populationRank, 15, 0)
       // Order DESCENDING (larger values win, Millbrea 40k wins over San Bruno 20k, both rank 7)
-      .thenByLog(population, 1000000000, 1, 100)
+      .thenByLog(population, 40000000, 1, 100)
       // Order ASCENDING (shorter strings are better than longer strings for map display and adds predictability)
       .thenByInt(name == null ? 0 : name.length(), 0, 31)
       .get();
@@ -55,15 +57,30 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
 
   private static final ZoomFunction<Number> LOCALITY_GRID_SIZE_ZOOM_FUNCTION =
     ZoomFunction.fromMaxZoomThresholds(Map.of(
-      6, 32,
-      7, 64
+      3, 24,
+      4, 24,
+      5, 24,
+      7, 24,
+      8, 32,
+      9, 32,
+      10, 32,
+      11, 24,
+      14, 24,
+      15, 16
     ), 0);
 
   private static final ZoomFunction<Number> LOCALITY_GRID_LIMIT_ZOOM_FUNCTION =
     ZoomFunction.fromMaxZoomThresholds(Map.of(
-      6, 8,
-      7, 6,
-      9, 4
+      3, 1,
+      4, 1,
+      5, 1,
+      6, 1,
+      8, 1,
+      9, 1,
+      10, 1,
+      11, 1,
+      14, 2,
+      15, 3
     ), 0);
 
   public void processOsm(SourceFeature sf, FeatureCollector features) {
@@ -148,7 +165,7 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           // This minZoom can be changed to smaller value in the NE data join step below
           minZoom = 11.0f;
           maxZoom = 15.0f;
-          kindRank = 3;
+          kindRank = 4;
           if (population == 0) {
             minZoom = 12.0f;
             population = 1000;
@@ -159,7 +176,7 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           // This minZoom can be changed to smaller value in the NE data join step below
           minZoom = 11.0f;
           maxZoom = 15.0f;
-          kindRank = 3;
+          kindRank = 5;
           if (population == 0) {
             minZoom = 12.0f;
             population = 200;
@@ -170,7 +187,7 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           // This minZoom can be changed to smaller value in the NE data join step below
           minZoom = 13.0f;
           maxZoom = 15.0f;
-          kindRank = 3;
+          kindRank = 6;
           if (population == 0) {
             minZoom = 14.0f;
             population = 100;
@@ -181,7 +198,7 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           // This minZoom can be changed to smaller value in the NE data join step below
           minZoom = 13.0f;
           maxZoom = 15.0f;
-          kindRank = 3;
+          kindRank = 7;
           if (population == 0) {
             minZoom = 14.0f;
             population = 50;
@@ -193,7 +210,7 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           // This minZoom can be changed to smaller value in the NE data join step below
           minZoom = 13.0f;
           maxZoom = 15.0f;
-          kindRank = 3;
+          kindRank = 8;
           if (population == 0) {
             minZoom = 14.0f;
             population = 1000;
@@ -204,19 +221,19 @@ public class Places implements ForwardingProfile.FeaturePostProcessor {
           kind = "neighbourhood";
           minZoom = 11.0f;
           maxZoom = 15.0f;
-          kindRank = 4;
+          kindRank = 9;
           break;
         case "quarter":
           kind = "macrohood";
           minZoom = 10.0f;
           maxZoom = 15.0f;
-          kindRank = 5;
+          kindRank = 10;
           break;
         case "neighbourhood":
           kind = "neighbourhood";
           minZoom = 12.0f;
           maxZoom = 15.0f;
-          kindRank = 6;
+          kindRank = 11;
           break;
       }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nvkelso** [comment]
Some screenshots before (white halo) and after (no halo). Some cities are labeled more than once in QGIS visualization because of the non-zero tile buffer.

Zoom 6 before
ca-zoom-6-before

Zoom 6 after
- arguably a few cities are being dropped that shouldn't be, but overall the effect is an improvement
ca-zoom-6-after

Zoom 7 before
ca-zoom-7-before

Zoom 7 after
ca-zoom-7-after

Zoom 8 before
ca-zoom-8-before

Zoom 8 after
ca-zoom-8-after


Zoom 9 before
ca-zoom-9-before

Zoom 9 after
ca-zoom-9-after


Zoom 10 before
ca-zoom-10-before

Zoom 10 after
ca-zoom-10-after

Zoom 11 before
ca-zoom-11-before

Zoom 11 after
ca-zoom-11-after

Zoom 12 before
ca-zoom-12-before

Zoom 12 after
ca-zoom-12-after

Zoom 13 before
ca-zoom-13-before

Zoom 13 after
ca-zoom-13-after

File: tiles/src/main/java/com/protomaps/basemap/layers/Places.java [hunk 1]
```diff
@@ -55,15 +57,30 @@ static int getSortKey(double minZoom, int kindRank, int populationRank, long pop
 
   private static final ZoomFunction<Number> LOCALITY_GRID_SIZE_ZOOM_FUNCTION =
     ZoomFunction.fromMaxZoomThresholds(Map.of(
-      6, 32,
-      7, 64
+      3, 24,
```
nvkelso, line 60: This GRID_SIZE sequence could be simplified, but needs to match the zoom steps in the GRID_LIMIT sequence below.

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -55,15 +57,30 @@ static int getSortKey(double minZoom, int kindRank, int populationRank, long pop
...
+      3, 1,
+      4, 1,
+      5, 1,
+      6, 1,
+      8, 1,
+      9, 1,
+      10, 1,
+      11, 1,
+      14, 2,
+      15, 3
```
nvkelso, line 83: You could argue the zoom 15 step could be a larger int, or not limited at all

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -287,12 +304,14 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
       //feat.setSortKey(minZoom * 1000 + 400 - populationRank * 200 + placeNumber.incrementAndGet());
       feat.setSortKey(getSortKey(minZoom, kindRank, populationRank, population, sf.getString("name")));
 
+      // This is only necessary when prepping for raster renderers
+      feat.setBufferPixels(16);
```
nvkelso, line 308: If Protomaps is only targeting vector renderers than this could be set to 0 to reduce file size

bdon: I think we still want a buffer for drawing townspots correctly very close to tile edges?

msbarry: If your label grid size isn't a factor of 256/power of 2 then you need a bit extra in order for label grid squares to be aware of points in the square from neighboring tiles otherwise the density might get off a bit around tile boundaries. Looks like 24 is the only problematic one. I think a buffer of 16 should be the minimum that works with label grid of 24.

If you want to go smaller you could either switch label grid to a power of 2 or use `FeatureMerge.removePointsOutsideBuffer` in `postProcess`

msbarry: I think the min buffer pixel limit should be `label grid size - gcd(256, label grid size)` - I should probably update planetiler to use that limit and post-filter to whatever you set here so users don't need to now or care about that implementation detail...

bdon:
@nvkelso here is a [Maperture example](https://stamen.github.io/maperture/#map=7.59/32.8784/-96.8376/0/0&showCollisions=true&showBoundaries=false&showDiff=false&viewMode=swipe&maps=%5B%7B%22name%22%3A%22Fetch%20URL%20at...%22%2C%22type%22%3A%22mapbox-gl%22%2C%22renderer%22%3A%22maplibre-gl%22%2C%22url%22%3A%22https%3A%2F%2Fbuild-metadata.protomaps.dev%2Fstyle%403.1.0%2Btheme%40light%2Btiles%4020240909.json%22%2C%22index%22%3A0%7D%2C%7B%22name%22%3A%22Fetch%20URL%20at...%22%2C%22type%22%3A%22mapbox-gl%22%2C%22renderer%22%3A%22maplibre-gl%22%2C%22url%22%3A%22https%3A%2F%2Fprotomaps.dev%2F~bdon%2Fplaces-thinning.json%22%2C%22index%22%3A1%7D%5D) with the default style (light).

It seems like the effect on the style is negligible, but tiles should be much lighter now. Other styles (like Mamata's) or raw unfiltered display in QGIS will have less places because their style filters based around minzoom are different. Is that the intended effect? Otherwise LGTM

bdon:
@nvkelso I may have mis-evaluated the above, how are you viewing zoom levels in QGIS? suspect that may not be consistent with either Leaflet/Tangram or MapLibre GL

bdon:
I'm using https://github.com/kgjenkins/qgis-zoom-level and I believe it displays Leaflet zoom levels (256x256 display pixels) and not maplibre / tile data ones (512x512 display pixels)

nvkelso:
> I'm using https://github.com/kgjenkins/qgis-zoom-level and I believe it displays Leaflet zoom levels (256x256 display pixels) and not maplibre / tile data ones (512x512 display pixels)

Same, 256px display pixels so Leaflet zooms.

nvkelso:
I need to do more testing to understand why the preview you linked to mostly looks the same...

bdon:
@nvkelso yes, I confirmed that it looks thinned out in QGIS. let me double check the tileset from that Maperture link was the right branch...

bdon:
@nvkelso the Maperture link has been updated - should be able to "view boxes" and visualize a significant improvement in hidden labels.

nvkelso:
@bdon New Mapeture [link](https://stamen.github.io/maperture/#map=8.01/37.5663/-122.2462/0/0&showCollisions=true&showBoundaries=false&showDiff=false&viewMode=mirror&maps=%5B%7B%22name%22%3A%22Fetch%20URL%20at...%22%2C%22type%22%3A%22mapbox-gl%22%2C%22renderer%22%3A%22maplibre-gl%22%2C%22url%22%3A%22https%3A%2F%2Fbuild-metadata.protomaps.dev%2Fstyle%403.1.0%2Btheme%40light%2Btiles%4020240909.json%22%2C%22index%22%3A0%7D%2C%7B%22name%22%3A%22Fetch%20URL%20at...%22%2C%22type%22%3A%22mapbox-gl%22%2C%22renderer%22%3A%22maplibre-gl%22%2C%22url%22%3A%22https%3A%2F%2Fgist.githubusercontent.com%2Fnvkelso%2F86e2a66baaf4f2aacb54d8865931025f%2Fraw%2F8f09d8adbcb32f7ae6d49cc44578a39933225196%2Fplace-thinning-v2.json%22%2C%22index%22%3A1%7D%5D) where I've shrunk the texts and icon padding. 

I tried floating the text around the townspot but that adds too many text labels for an apples to apples comparison, and shows poorly in dense areas with long names like around New York city. 

Adding sort-key solved the flickering in New York and Barcelona etc. But to also solve in Delhi where the min_zoom is correct but the population_rank is wrong, you'd want to modify this to use the `min_zoom` value first (multiplied by 100) and then added to `population_rank` and then multiplied by -1 so reverse sort it, though to solve the Delhi problem.

Something like (see src though):

```
        "symbol-sort-key": [
          "*",
          -1,
          [
            "get",
            "population_rank"
          ]
        ],
        "text-padding": [
          "interpolate",
          [
            "linear"
          ],
          [
            "zoom"
          ],
          5,
          2,
          8,
          2,
          12,
          2
        ],
        "icon-padding": [
          "interpolate",
          [
            "linear"
          ],
          [
            "zoom"
          ],
          0,
          0,
          8,
          2,
          10,
          2,
          12,
          2,
          22,
          2
        ],
        "text-variable-anchor": {
          "stops": [
            [
              7,
              [
                "top",
                "top",
                "left",
                "right"
              ]
            ],
            [8, ["center"]]
          ]
        },
```
A few before and afters:

image

image

image

nvkelso:
@msbarry is there an easy way to limit the grid thinning to certain zooms? I'd prefer it not start until say zoom 7.

msbarry:
Yeah

> @msbarry is there an easy way to limit the grid thinning to certain zooms? I'd prefer it not start until say zoom 7.

You mean you want at zoom <= 7 right?  There's a `feature.setPointLabelGridSizeAndLimit(int maxzoom, double size, int limit)` method that limits label grid filtering to <= maxzoom.

nvkelso:
> You mean you want at zoom <= 7 right? There's a feature.setPointLabelGridSizeAndLimit(int maxzoom, double size, int limit) method that limits label grid filtering to <= maxzoom.

@msbarry No, zoom > 7 as the early zooms are highly curated and it's OK for say Oakland and San Francisco to appear next to each other in the same collision grid cell.

msbarry:
ah OK so if you want to apply to zooms >= 7 you can do something like `feature.setPointLabelGridSize(ZoomFunction.minZoom(7, size)).setPointLabelGridLimit(ZoomFunction.minZoom(7, limit))`

wipfli:
I have ported your code over to #437, closing...

PR #308, opened 2024-09-29
Add 2nd language option

wipfli:
Adds an option to display labels in a second target language.

image

The code for regular language localization in `get_multiline_name()` takes the target script and the local script(s) into account. Since there is already a high level of complexity in this function I decided to create a new function `get_dual_name(langPrimary, langSecondary)`. It roughly behaves like this:

```
if both target languages are present
  if both languages are the same
    only show one language
  else
     show both languages
else if no target language is present
  use default "name"
else
  show the language which is present
``` 

When refactoring the code to support a second language, I noticed that the optional `script` argument in `get_multiline_name()` does not make a lot of sense. I currently don't see a situation where the user would want to specify a script for a language other than the defaults that we have. I therefore removed the script parameter. From semver's point of view this is a breaking change.

### DIFF
--- a/app/src/MapViewComponent.tsx
+++ b/app/src/MapViewComponent.tsx
@@ -70,6 +70,7 @@ export const isValidPMTiles = (tiles?: string): boolean => {
 function getMaplibreStyle(
   theme: string,
   lang: string,
+  langSecondary: string,
   localSprites: boolean,
   tiles?: string,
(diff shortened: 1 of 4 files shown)


wipfli:
Closing in favor of https://github.com/protomaps/docs/pull/66

Issue #303, opened 2024-09-23
final touches on 4.0 breaking changes


bdon:
@nvkelso

* fold `kind=medium_road` into `kind=major_road`: [Tilezen roads](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
* remove non-integer `kind_detail` from `boundaries`. remove `min_admin_level` and fold into `kind_detail`. [tilezen boundaries](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#boundaries-properties-common)
* rename `ele` to `elevation` in POIs [tilezen pois](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#poi-properties-only-on-kindpeak-and-kindvolcano)
* remove integer `level` from `roads` and replace with `is_bridge=true` or `is_tunnel=true` [tilezen roads](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#poi-properties-only-on-kindpeak-and-kindvolcano)

Note: Tangram relies on `sort_rank` for proper layering which is not possible with MapLibre GL, so maybe we want to keep integer `level`?

Still working through one more issue with `kind=rail` appearing in `roads` layer.

nvkelso:
> Note: Tangram relies on sort_rank for proper layering which is not possible with MapLibre GL, so maybe we want to keep integer level?

Yes, ok to keep `level` to allow semi-complicated over/under road layering. Though for perf and simplicity these are often flattened to just 3 sets of layers (any tunnels of a given kind, any surface of a given kind, any bridges of a given kind).

bdon:
I think we need to migrate all current `rail=` (physical rail) into `roads` (transportation), which shows up quite late (~z13), and then leave `transit` empty. `transit` in a later 4.x release will be abstract services (Subway, HSR, etc) that appears at lower zoom levels.

@nvkelso let me know if this is consistent with your design, as well as `kind_detail` in boundaries above?

nvkelso:
Yes, sounds good

bdon:
Closing this one out. Will address `level` likely via the server-assisted approach in #72

PR #309, opened 2024-10-01
Tilezen tags [#303]

bdon:


tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -24,7 +24,6 @@ public String name() {
   public void processNe(SourceFeature sf, FeatureCollector features) {
     var sourceLayer = sf.getSourceLayer();
     var kind = "";
-    var kindDetail = "";
```
nvkelso, line 27: `kindDetail` was implemented in v3 as a forward compatible change to Tilezen 1.9 schema. It's not necessary to remove it v4, although you can if you want. Natural Earth has a lot of extra fidelity that wasn't captured before... but for rendering probably just kind and the disputed boolean are sufficient for most basic maps.

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -146,9 +136,9 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
         .setAttr("kind", kind)
         // Don't label lines to reduce file size (and they aren't shown in styles anyhow)
         //.setAttr("name", sf.getString("name"))
-        // Preview v4 schema (disabled)
         //.setAttr("min_zoom", sf.getLong("min_zoom"))
-        .setAttr("min_admin_level", adminLevel)
+        .setAttr("kind_detail", adminLevel)
+        .setAttr("sort_rank", 289)
```
nvkelso, line 141: This is a coarse approximation of the varied sort_rank from Tilezen.

tiles/src/main/java/com/protomaps/basemap/layers/Boundaries.java
```diff
@@ -192,7 +176,6 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
         switch (minAdminLevel.getAsInt()) {
           case 2 -> {
             kind = "country";
-            kindDetail = "2";
```
nvkelso, line 195: Be wary of removing `kindDetail` here as `kind` alone doesn't capture some of the per country variance –– right now Protomaps v3 and v4 would push _more_ boundaries features from OSM into tiles than Tilezen v1.9 schema, and the `kindDetail` would provide a way to filter those out in the style. The better fix would be to be more selective at tile cut, of course.

tiles/src/main/java/com/protomaps/basemap/layers/Pois.java
```diff
@@ -456,7 +456,7 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
           // Core OSM tags for different kinds of places
           // Special airport only tag (to indicate if it's an airport with regular commercial flights)
           .setAttr("iata", sf.getString("iata"))
-          .setAttr("ele", sf.getString("ele"))
+          .setAttr("elevation", sf.getString("ele"))
```
nvkelso, line 459: good

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -134,23 +133,20 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
...
         .setId(FeatureId.create(sf))
-        // Core Tilezen schema properties
         .setAttr("kind", kind)
         // To power better client label collisions
         .setAttr("min_zoom", minZoom + 1)
         .setAttrWithMinzoom("ref", shield.text(), minZoomShieldText)
         .setAttrWithMinzoom("shield_text_length", shieldTextLength, minZoomShieldText)
         .setAttrWithMinzoom("network", shield.network(), minZoomShieldText)
-        // Core OSM tags for different kinds of places
-        .setAttrWithMinzoom("layer", Parse.parseIntOrNull(sf.getString("layer")), 12)
```
nvkelso, line 145: Tilezen baked the layer into sort_rank calculation... but we're using 400 everywhere in this PR so there would be information loss. Without being more sophisticated about the sort_rank, the more interesting highway junctions won't render correctly anymore.

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
```diff
@@ -164,24 +160,117 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
...
       }
 
       // Server sort features so client label collisions are pre-sorted
       feat.setSortKey(minZoom);
 
       OsmNames.setOsmNames(feat, sf, minZoomNames);
+    } // end highway=
+
+    // non-highway features
+    // todo: exclude railway stations, levels
```
nvkelso, line 181: This is important, there were several Tilezen versions where we didn't have that filtering and the rendering was borked.

tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
bdon: This first PR is just go get the key in there so that Tangram renders, we will add a more detailed sort rank later

nvkelso: Sounds good!

PR #313, opened 2024-10-09
set text-justify: auto on localities that might use a townspot symbol.

bdon:
@wipfli 

CleanShot 2024-10-09 at 14 31 33@2x

### DIFF
--- a/app/src/MapViewComponent.tsx
+++ b/app/src/MapViewComponent.tsx
@@ -18,7 +18,7 @@ import { language_script_pairs } from "../../styles/src/language.ts";
 import { LayerSpecification } from "@maplibre/maplibre-gl-style-spec";
 
 import { FileSource, PMTiles, Protocol } from "pmtiles";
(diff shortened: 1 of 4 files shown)


wipfli:
Looks good to me, thanks!

Just to make sure that everyone is aware of the following detail: When a label consists of multiple scripts, e.g. when you set the target language to Greek and you look at Switzerland where all labels are Latin, then 

* if `name:el` is present, it will show a label with two lines, e.g. `Ζυρίχη\nZürich`
* if `name:el` is not present, it will show an empty line and then the local name, e.g. `\nSchaffhausen`

So the symbolizer is either to the left or to the top left of the text. If think that looks good but it is something one needs to be aware of...

image

bdon:
I would prefer the text is correctly aligned when a single line, and was looking for ways to fix it; it seems like we need a way to make the newline conditional, or somehow move the newline into the case statement of the second line?

Issue #312, opened 2024-10-04
Ferry line labels


wipfli:
**Location**
Geneva high zoom

https://maps.protomaps.com/#map=15.92/46.206564/6.149963&theme=light&lang=en&tiles=https://build.protomaps.com/20241004.pmtiles

**Screenshots**
![Screenshot_20241004-194357](https://github.com/user-attachments/assets/1a261832-ad0e-42d3-b346-dad5464789b1)

Are labels on ferry lines on purpose?

nvkelso:
Doh! Ferry line labels should be excluded from early- and mid-zooms as they confuse most people (they are often long strings of text, and make it look like there is a road in the water when it's a ferry instead). 

Ferry line labels are optional to taste somewhere between zooms 16 and 18. I'd recommend 18 for Protomaps. What to file a PR? They should also use a different text treatment than road labels, perhaps blue italic text?

bdon:
ferry lines seem to be getting `kind=other` and thus labeled because it's looking for `ferry=*` in OSM: https://github.com/protomaps/basemaps/blob/main/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java#L228

This seems like a mistake relative to Tilezen v1.9, will open a PR

bdon:
Fixed in 4.0.1

PR #314, opened 2024-10-09
roads kind=ferry: filter by route=ferry instead of ferry=*. fixes [#312]

bdon:
@nvkelso this is to match https://tilezen.readthedocs.io/en/latest/layers/#road-transportation-kind_detail-values-and-zoom-ranges

> Ferries from both Natural Earth and OpenStreetMap are shown starting at zoom 5 with kind values of ferry.

as well as what's in http://tangrams.github.io/bubble-wrap/#15.595833333333331/37.7923/-122.3811


nvkelso:
🎉

PR #324, opened 2024-10-19
fix typo `min_height` attributes in buildings layer

### LINKED ISSUE #323: no Min Height in buildings layer
I want to show 3D building from `buildings` layer, and for some buildings we need `min_height`. here the result:

image

for comparison from [openfreemap](https://openfreemap.org/):

image

```json
{
      "id": "buildings-3d",
      "type": "fill-extrusion",
      "source": "protomaps",
      "source-layer": "buildings",
      "minzoom": 14,
      "paint": {
        "fill-extrusion-base": [
          "get",
          "min_height"
        ],
        "fill-extrusion-color": "hsl(35,8%,85%)",
        "fill-extrusion-height": [
          "get",
          "height"
        ],
        "fill-extrusion-opacity": 1
      }
    },
```

I am not familiar with layers generation, but is this line correct?

https://github.com/protomaps/basemaps/blob/f7bcb4869434030ba23271f43e9dbcf6bfc5d63f/tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java#L20

JfrAziz:
close #323

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
@@ -17,7 +17,7 @@ import java.util.regex.Pattern;
 public class Buildings implements ForwardingProfile.LayerPostProcesser {
 
   static final String HEIGHT_KEY = "height";
-  static final String MIN_HEIGHT_KEY = "height";
+  static final String MIN_HEIGHT_KEY = "min_height";
 
   @Override
   public String name() {


nvkelso (commented):
LGTM, thanks!

nvkelso:
The test may need to be updated, too: 

- https://github.com/protomaps/basemaps/blob/main/tiles/src/test/java/com/protomaps/basemap/layers/BuildingsTest.java#L48

Though the efficacy of the existing test isn't clear to me.

nvkelso:
Also worth noting in the screenshot @bdon we may want to add a minimum default building height when the height is null in the style for those hospital buildings to the right of the London Eye that don't pop in Protomaps but do in the other project's style (OpenMapTile based).

bdon:
Thanks!

Issue #44, opened 2023-06-12
Incorporate country-coder for locale-specific logic


bdon:
https://github.com/rapideditor/country-coder provides a complete, fast and purely geometric dataset we can use in-memory for java to determine locale for cartographic logic.

1ec5:
Some use cases for this functionality, off the top of my head:

* Bank and ATM icons based on the local currency symbol
* Culturally sensitive icons for Buddhist temples and post offices in Japan that would be poorly understood elsewhere (to say the least)
* Regionally appropriate glosses of “local” names, independently of OSM’s `name=*` tags (which sometimes reflect some compromises)
* Romanizations of non-Latin names according to national standards
* Prioritizing each country’s major road network at low zoom levels independently of `highway=trunk` tagging, which can be a mess
* [One-way icons offset from the road](https://github.com/osm-americana/openstreetmap-americana/pull/995#discussion_r1545868242), as in conventional print maps, but on the correct driving side
* Route shields for some route networks like the Asian Highway that vary dramatically from country to country
* The UK _really_, _really_ likes their roads color-coded just so based on highway classifications

bdon:
Added in #328

PR #328, opened 2024-10-24
add country-coder json [#44]

bdon:
@wipfli @nvkelso


tiles/src/main/java/com/protomaps/basemap/feature/CountryCoder.java
```diff
@@ -0,0 +1,84 @@
...
+import com.onthegomap.planetiler.geo.GeoUtils;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.locationtech.jts.geom.*;
+import org.locationtech.jts.index.strtree.STRtree;
+
+public class CountryCoder {
```
wipfli, line 15: You might want to use Planetiler's `PolygonIndex`, see https://github.com/onthegomap/planetiler/blob/f5fe38a4d4400f3aa30998a1eb3ac551ce3e47d9/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/PolygonIndex.java#L20, cc @msbarry

bdon: We may need to enhance that to handle (Polygon contains LineString) - looks like Contains only supports Point and Intersects works for all geometries? TBD if it actually matters

msbarry:
FYI I'm planning to add something very similar to planetiler soon to support name abbreviations and other use cases like this: https://github.com/onthegomap/planetiler/issues/14#issuecomment-2352691236

bdon:
@msbarry great, I could upstream this implementation or merge it with PolygonIndex once it's proven out, unless you see any major things missing? Biggest TBD feels like what happens at boundaries

tiles/src/main/java/com/protomaps/basemap/feature/CountryCoder.java
msbarry: I was thinking about that, it seems like country coder should return the country that feature is "most within" so we could sample a few points? Here's the api that prepared geometry exposes when testing geometries: https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/prep/PreparedGeometry.html

msbarry:
> @msbarry great, I could upstream this implementation or merge it with PolygonIndex once it's proven out, unless you see any major things missing? Biggest TBD feels like what happens at boundaries

That would be great! My plan was to eventually have configurable backends that require downloading different data sources (natural earth borders by POV or overture divisions) but the fallback would be to use this country-coder json source bundled in the jar.

PR #330, opened 2024-11-04
Change `kind` attribute of OSMData sourced-water to `ocean`

### LINKED ISSUE #329: water layer: kind=ocean for zooms > 5 rather than changing to kind=water
**Description**
The value of `kind` for oceans switches from `ocean` at and below z5, to `water` at and above z6. I would find it really helpful if this could stay consistent across zoom levels. I was trying to extract just the oceans from the protomaps builds and mashup with a different data source for inland water, but I can't do this beyond z5 without an attribute to differentiate the oceans (in v4; in v3, I could use the presence of the `natural` key to do this). 

As a workaround, I'm going to try the polygons from [OSMData](https://osmdata.openstreetmap.de/data/water-polygons.html) as an ocean source, but it would be a great enhancement if this wasn't necessary. I can imagine other use cases where it would be helpful to differentiate oceans from inland water—someone might want different colors for each, for example.

[Example zoomed right before the changover](https://pmtiles.io/?url=https%3A%2F%2Fbuild.protomaps.com%2F20241104.pmtiles#map=5.83/36.861/-73.877)

dericke:
Closes #329 

Let me preface by saying I don't know Java and I'm not very familiar with this codebase, so please excuse any amateur mistakes or oversights. I've marked this PR as a draft to ensure nobody mistakes me for someone who knows what they're doing.

As described in #329, the value of `kind` for oceanic features switches from `ocean` in z5 to `water` in z6. I think this inconsistency at different zoom levels is undesirable, and it precludes data users from being able to treat oceans differently from inland water. For example, in my own case, I'm wanting to use inland water data from the USGS rather than OSM for a US-focused map, while still using Natural Earth/OSMData for the oceans, but need a way to differentiate the two by attribute.

Since the water from the OSMData source only includes oceans (by design), it should be accurate to tag all water features from this source as `kind: ocean`, which is what this PR does.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Water.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Water.java
@@ -25,7 +25,7 @@ public class Water implements ForwardingProfile.LayerPostProcesser {
 
   public void processPreparedOsm(SourceFeature ignoredSf, FeatureCollector features) {
     features.polygon(this.name())
-      .setAttr("kind", "water")
+      .setAttr("kind", "ocean")
       .setAttr("sort_rank", 200)
       .setZoomRange(6, 15).setBufferPixels(8);
   }


bdon:
Good catch, thanks!

Issue #335, opened 2024-12-06
Missing building numbers?


SiarheiFedartsou:
**Description**
Hey, is it expected that Basemap doesn’t include building numbers?

**Screenshots**
The same place on openstreetmap.org and Basemap
![2E521F24-7025-44F1-9B44-2DB6CC2F56D2](https://github.com/user-attachments/assets/79c2b309-cce9-465f-9d89-cb5417efdecc)
![6FE98F34-AB3A-44B1-86D4-C7415BCA25FB](https://github.com/user-attachments/assets/7f554cb9-2095-451e-9779-b42322452149)

SiarheiFedartsou:
https://github.com/protomaps/basemaps/pull/336

pietervdvn:
The main discussion for this is in #280 , this can be considered a duplicate issue and (more or less) resolved.

bdon:
Resolved by #380 #387

PR #336, opened 2024-12-08
Add house numbers layer

SiarheiFedartsou:
### Description
This is an attempt to add house numbers to base maps.

Since I am not sure in approach, I decided to create this PR with preliminary implementation first and decide on the right approach.

How it looks like now in Monaco:
Screenshot 2024-12-08 at 13 21 44

Open questions/TODO:
1. Design (these red itailic labels is definitely not what we want + we need to decide on zoom levels etc)
2. In OSM `addr:housenumber` is usually:
 - tag on separate node
 - tag on `building` way
 - both of above 
Current implementation handles only separate nodes for now. To handle ways as well we probably need a kind of a deduplication logic to not show the same housenumber twice (at glance we can try matching by other `addr:*` tags to identify duplicates)

### Issue
https://github.com/protomaps/basemaps/issues/335


tiles/src/main/java/com/protomaps/basemap/layers/Housenumbers.java
```diff
@@ -0,0 +1,34 @@
...
+import com.onthegomap.planetiler.geo.GeometryException;
+import com.onthegomap.planetiler.reader.SourceFeature;
+import com.protomaps.basemap.feature.FeatureId;
+
+import java.util.List;
+
+public class Housenumbers  implements ForwardingProfile.LayerPostProcesser {
+  public void processOsm(SourceFeature sf, FeatureCollector features) {
+    if (sf.hasTag("addr:housenumber")) {
+      if (sf.isPoint()) {
```
wipfli, line 15: Also a lot of ways in osm have a house number: https://taginfo.openstreetmap.org/keys/addr%3Ahousenumber#overview

SiarheiFedartsou: Yep, I agree. As I mentioned in PR description I wanted to handle them as well, but wanted to first agree on the general approach. Do you think such implementation does have a chance to be eventually merged? 😀 Ofc after handling all the cases, adding tests etc...

wipfli: I think it is a good idea to add house numbers.

bdon: This looks like a great start! We should align the layer name and properties to conform with the Tilezen docs: https://tilezen.readthedocs.io/en/latest/layers/#buildings-and-addresses

so it should be points included in the `buildings` layer with a property called `addr_housenumber`

tiles/src/main/java/com/protomaps/basemap/layers/Housenumbers.java
```diff
@@ -0,0 +1,34 @@
...
+        features.point(this.name())
+          .setId(FeatureId.create(sf))
+          .setAttr("housenumber", sf.getString("addr:housenumber"))
+          .setZoomRange(10, 17);
```
bdon, line 19: We should limit the house numbers to exist only in the zoom 15 tiles, in order to not bloat the tile size. A visual reference we use for deciding what data goes in what zoom is http://tangrams.github.io/bubble-wrap - house numbers only appear above zoom ~18 which means we can put them in our tileset at 15 and define the style to appear > 18, etc. Is that zoom level good enough for your use case?

SiarheiFedartsou: Yes, I think so :)

SiarheiFedartsou:
Hey @bdon !

Thanks for your feedback. 
I moved house numbers to buildings layer as you suggested and also made some fixes related to style, zoom level etc. 

Some questions:
1. Did I understand your suggestions correctly? 😄 I mean I want to first check if I am going the right way and if it is so I could add tests and may be some other things to get it eventually merged.
2. Do we need `addr_street` to be present in tiles? It is in Tilezen's spec, but I don't use it in style now, so I drop it to not bloat resulting tiles https://tilezen.readthedocs.io/en/latest/layers/#buildings-and-addresses

styles/src/themes.ts
```diff
@@ -188,6 +191,9 @@ export const CONTRAST: Theme = {
   state_label: "#777777",
   state_label_halo: "#ffffff",
   country_label: "#9590aa",
+
+  housenumbers_label: "#91888b",
```
SiarheiFedartsou, line 195: I used the same values as `roads_label_minor` for all themes... Looks okay on my taste...
Screenshot 2024-12-14 at 09 53 14

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -93,13 +94,46 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
...
+        if (item.tags().containsKey("addr_housenumber")) {
+          groupedAddresses.computeIfAbsent(item.tags(), k -> new ArrayList<>()).add(item);
+        } else {
+          buildings.add(item);
+        }
+      }
+
+      for (var address : groupedAddresses.values()) {
+        var feature = address.get(0);
+        feature.tags().remove("addr_street");
```
SiarheiFedartsou, line 132: I drop `addr_street` here as I don't use it in style - only for deduplication here.

styles/src/base_layers.ts
```diff
@@ -594,6 +594,7 @@ export function nolabels_layers(
       type: "fill",
       source: source,
       "source-layer": "buildings",
+      filter: ["!=", "kind", "address"],
```
bdon, line 597: I think this should be an opt-in instead of an opt-out, e.g. `kind` in `building` or `building_part`

styles/src/themes.ts
```diff
@@ -188,6 +191,9 @@ export const CONTRAST: Theme = {
...
+  house_numbers_label_halo: "#ffffff",
```
bdon, line 196: maybe `housenumbers` one word instead of with `_`?

bdon:
> Do we need addr_street to be present in tiles? It is in Tilezen's spec, but I don't use it in style now, so I drop it to not bloat resulting tiles

No, we don't need it if it's not used in the default style. Removing bloat is great.

tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
```diff
@@ -93,13 +98,46 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
...
   }
 
   @Override
   public List<VectorTile.Feature> postProcess(int zoom, List<VectorTile.Feature> items) throws GeometryException {
     if (zoom == 15) {
-      return items;
+      List<VectorTile.Feature> buildings = new ArrayList<>();
+
+      // deduplicate addresses
+      HashMap<Map<String, Object>, List<VectorTile.Feature>> groupedAddresses = new LinkedHashMap<>();
```
bdon, line 125: Is this meant to address one building that is labeled twice on the map? Is there a specific place in the world where that problem is pervasive or is it an edge case where maybe it's a data problem in OSM?

bdon: NVM, I see the issue is that it also includes POIs that have an `addr:housenumber` in OSM. I think for the first pass we should limit the house number displays to just those tagged on buildings and not POIs, because de-duplicating them in post-process is going to lead to non-deterministic results potentially, and ideally POIs will be labeled by an icon symbology instead of just a housenumber. What do you think? Labeling on OSM Carto seems to be focused on the building `addr:housenumber` too.

do-me:
Just saw this PR. A month ago, I was also digging into house numbers and used a more basic style, after @wipfli gave me some hints in https://github.com/protomaps/PMTiles/issues/503. This is the repo to reproduce this style: https://github.com/do-me/basemap-starter.

image

I too simply used `.setAttr("house_number", sf.getTag("addr:housenumber"))` in basemaps but found - as I mentioned [here](https://github.com/protomaps/PMTiles/issues/503#issuecomment-2520022240) and @wipfli pointed to [here](https://github.com/protomaps/basemaps/pull/336#discussion_r1875036213) - that house number tagging across different countries is very inconsistent. Best case is house numbers attached to buildings, but in some cases it's a single point (I think it was France) or added elsewhere.

I just created a world pmtiles file based on the above config and the result is **very** heterogenous. I can share it if you like.

Really good in Zurich: 

image

Very inconsistent in Munich: 

image

For comparison, see [OSM](https://www.openstreetmap.org/node/315756947#map=17/48.150215/11.641924), where house numbers do exist, just not on the buildings: 

image

Practically no house numbers in France: 

image

So @bdon I would highly encourage to use all above mentioned tags right from the start and not only rely on buildings.

Also, there are many open discussions/issues scattered in protomaps discussions, basemaps & pmtiles repo about house numbers, so it might make sense to link these threads better for people see that development is going on here.

bdon:
> all above mentioned tags right from the start and not only rely on buildings.

Are you suggesting using POIs for addresses? It would be helpful to understand how the OSM Carto screenshot you show above derives its more detailed numbers. I mentioned in the PR comment that using POIs for addresses creates lots of duplicate address points and we do not have a clear way to de-duplicate them in a deterministic way.

bdon:
A more complex option is performing a point-in-polygon test on every point to see if its `addr:housenumber` matches the building that contains it, and then throwing it away if it matches. 

If the building does not have a house number, but multiple points have identical house numbers, those will still show as duplicates in the final map.

bdon:
Looking at OSM as an example: https://www.openstreetmap.org/node/3046828339

what if we included point `addr:housenumber` only if there is no non-`addr:` tags? This would exclude all POIs. I will experiment with this on a new build.

klokan:
CC-BY

bdon:
FYI the PR as-is adds about ~1GB to the total planet tileset, which is reasonable. It looks like the current logic chooses the "first" one that makes it into the tileset is based on sort rank and seems deterministic, so I was wrong about that. Running it again to check.

do-me:
> FYI the PR as-is adds about ~1GB to the total planet tileset, which is reasonable. It looks like the current logic chooses the "first" one that makes it into the tileset is based on sort rank and seems deterministic, so I was wrong about that. Running it again to check.

So did you decide to merge this PR in the next release? If so, I would wait for it instead of using my own pipeline (https://github.com/do-me/basemap-starter) :)

bdon:
I merged it in #380 after doing some style cleanup and adding tests.

I think there is some broken conditions right now like if the address is only tagged on the building centroid, we can handle that in a follow up issue. Thanks!

PR #341, opened 2025-01-04
Fix geometry-type filter on water to be MapLibre v5 compatible.

bdon:
* Replace [geometry-type] with  to filter geometries correctly.


wipfli (approved):
works with 4.7.1 and 5.0.0

bdon:
Released in protomaps-themes-base 4.3.1

Issue #337, opened 2024-12-26
Antarctica Landcover does not extend to bottom of earth


piemadd:
**Location**
Include a link to a public URL with coordinates in the hash and zoom level.

**Screenshots**
Can be viewed on the build viewer: [https://maps.protomaps.com/#tiles=https://build.protomaps.com/20241223.pmtiles&map=0.95/0/-73.4&theme=dark&lang=en&show_boxes=true](https://maps.protomaps.com/#tiles=https://build.protomaps.com/20241223.pmtiles&map=0.95/0/-73.4&theme=dark&lang=en&show_boxes=true)
![image](https://github.com/user-attachments/assets/438ce810-a72b-48b8-97eb-eec0514eba43)

Is also more noticeable within maplibre's experimental globe rendering
![image](https://github.com/user-attachments/assets/b4192804-b755-4d91-8407-cf51c4027064)

**Required information**
 - Browsers: Firefox 133.0.3 and Chromium 131.0.6778.85 snap
 - Renderer versions: MapLibre GL JS 4.7.1, 5.0.0-pre.6, and 5.0.0-pre.10
 - Tileset version 20241223.pmtiles and 202412226 (from [https://maps.protomaps.com/builds/](https://maps.protomaps.com/builds/))
 - Style JSON: standard dark style on the preview

This seems to be an issue with the data, not the style. If I have some free time I could dive into the that, but its been a few years since I've touched Java seriously.

piemadd:
Going back it seems the "issue" started sometime between 3.4.1 and 3.5.1, which seems to be from the addition of the landcover layer and not from the removal of something else. This can be seen in the [visual tests](https://maps.protomaps.com/visualtests/?leftTiles=https://build.protomaps.com/20240401.pmtiles&rightTiles=https://build.protomaps.com/20240408.pmtiles) along with comparing [20240105](https://pmtiles.io/?url=https%3A%2F%2Fpm.transitstat.us%2F20241226.pmtiles#map=0.93/0/-6) with [20241226](https://pmtiles.io/?url=https%3A%2F%2Fpm.transitstat.us%2F20241226.pmtiles#map=0.93/0/-6).

![image](https://github.com/user-attachments/assets/611c3fd3-c6dd-4975-9649-bdefb9712550)

wipfli:
I think it is because the landcover dataset does not extend to below some latitude in the south. Maybe this could be fixed in the source data?

wipfli:
[...]

PR #342, opened 2025-01-04
Fix Antarctica landcover

wipfli:
Fixes https://github.com/protomaps/basemaps/issues/337

https://maps.protomaps.com/#map=0.97/-10.8/-17.6&theme=light&lang=en&tiles=https://protomaps.dev/~wipfli/antarctica.pmtiles

image

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
@@ -30,9 +30,13 @@ public class Earth implements ForwardingProfile.LayerPostProcesser {
     } else if (sourceLayer.equals("ne_10m_land")) {
       features.polygon(this.name()).setZoomRange(5, 5).setBufferPixels(8).setAttr("kind", "earth");
     }
+    // Daylight landcover uses ESA WorldCover which only goes to a latitude of roughly 80 deg S.
+    // Parts of Antarctica therefore get no landcover = glacier from Daylight.
+    // To fix this, we add glaciated areas from Natural Earth in Antarctica.
     if (sourceLayer.equals("ne_10m_glaciated_areas")) {
       try {
         Point centroid = (Point) sf.centroid();
+        // Web Mercator Y = 0.7 is roughly 60 deg South, i.e., Antarctica.
         if (centroid.getY() > 0.7) {
           features.polygon("landcover")
             .setAttr("kind", "glacier")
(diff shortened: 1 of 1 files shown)


nvkelso:
Bonus points if you also fix up northern Greenland, north of around 82.8° N, the other common area people complain about:

image

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
```diff
@@ -29,6 +30,19 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
     } else if (sourceLayer.equals("ne_10m_land")) {
       features.polygon(this.name()).setZoomRange(5, 5).setBufferPixels(8).setAttr("kind", "earth");
     }
+    if (sourceLayer.equals("ne_10m_glaciated_areas")) {
```
nvkelso, line 33: This section deserves a comment explaining the problem and why this particular solution works

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
```diff
@@ -29,6 +30,19 @@ public void processNe(SourceFeature sf, FeatureCollector features) {
...
+      try {
+        Point centroid = (Point) sf.centroid();
+        if (centroid.getY() > 0.7) {
```
nvkelso, line 36: This works, but it's confusing this isn't a negative number, and 70 would normally not result in the northern tip of Chilean Antarctica (many international bases there) showing up correctly as ice.

Please add a comment explaining that you're targeting ice features in Antarctica south of -60°.

nvkelso (approved):
Nits about comments and request to also fix northern Greenland.

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
wipfli: Added one, thanks for the tip!

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
wipfli: Web Mercator Y in Planetiler goes from 0 to 1 from 85 deg N to 85 deg S. The equator is at Web Mercator Y = 0.5

wipfli:
I tried Greenland with something like ` y > 0.7 || y < 0.3` but it did not have a big effect. Most of these areas close to the water that are green are not covered by glacieated NE areas.

tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
wipfli: Added the comment

nvkelso:
Comments look good!

For Greenland, make a fake rectangle of glacier in just that area? That would look better than the grey. But a nit, can merge this as-is.

Issue #51, opened 2023-06-24
Clipping to AOI


bdon:
Instead of all global (OSM Water, NE) features appearing at 0,0,0, every tile in a non-planet build should be intersected with a passed AOI. The AOI can be passed on the command line via GeoJSON.

I believe this is possible using only the public Planetiler APIs like `postProcess`, but it will require some tricks to make it efficient.

bdon:
Proof of concept is now available here: (file may expire):

https://maps.protomaps.com/#theme=light&lang=en&tiles=https://protomaps.dev/~bdon/switzerland.pmtiles&map=7.31/46.823/8.225
![CleanShot 2025-01-08 at 17 43 44@2x](https://github.com/user-attachments/assets/7bf4bd23-c8b7-453b-a9d0-b967d09d2045)

This is a good feature for national-level maps of small countries, as it avoids pulling in too much outside data and is visually cleaner. TODO make it work for multi polygons and ensure it scales up to very large AOIs (all of the US with a detailed boundary, for example)

bdon:
Implemented in tiles 4.2.0

PR #351, opened 2025-01-09
Clip tileset to a polygonal boundary

bdon:
Implements #51 

This is to address making non-global maps: for example, a map of a single continent, country, or city. The existing tooling has a path to doing this with `pmtiles extract` on a planet file, but this has a downside: the output tileset is not clipped to the AOI (area of interest) and shows the entire detail of the original source tileset at lower zoom levels. 

Additionally, if the input OSM data has features that lie outside of the AOI, like oceans or roads, those extend beyond the boundary. This is not a fatal problem for many use cases, but we can do better in our tile generation program by clipping every feature to the AOI.

The challenges here are:

* Making this fast: we need a efficient spatial representation of the clipping area, because running a constructive polygon operation like clip on 100% features is extremely slow
* Robustness: constructive geometry operations can create self-intersections
* MultiPolygon support: we need to support polygons with holes and multi polygons: for example, South Africa should not include Lesotho, and France should include overseas departments without pulling in lots of extra data.


tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+  private final Map<Integer, TiledGeometry.CoveredTiles> coverings;
+  private final Stats stats;
+
+  public Clip(Stats stats, Geometry input) {
+    this.stats = stats;
+    var clipGeometry = latLonToWorldCoords(input).buffer(0.00001);
+    tiledGeometries = new HashMap<>();
+    coverings = new HashMap<>();
+    try {
+      for (var i = 0; i <= 15; i++) {
```
bdon, line 37: Need to get min/max from the actual program somewhere.

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+import org.locationtech.jts.operation.overlayng.OverlayNGRobust;
+
+public class Clip implements ForwardingProfile.TilePostProcessor {
+  private final Map<Integer, Map<TileCoord, List<List<CoordinateSequence>>>> tiledGeometries;
+  private final Map<Integer, TiledGeometry.CoveredTiles> coverings;
+  private final Stats stats;
+
+  public Clip(Stats stats, Geometry input) {
+    this.stats = stats;
+    var clipGeometry = latLonToWorldCoords(input).buffer(0.00001);
```
bdon, line 33: Maybe this should be configurable buffer size via arguments?

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+    tiledGeometries = new HashMap<>();
+    coverings = new HashMap<>();
+    try {
+      for (var i = 0; i <= 15; i++) {
+        var extents = TileExtents.computeFromWorldBounds(i, WORLD_BOUNDS);
+        double scale = 1 << i;
+        Geometry scaled = AffineTransformation.scaleInstance(scale, scale).transform(clipGeometry);
+//        var simplified = DouglasPeuckerSimplifier.simplify(scaled, 0.25/256);
```
bdon, line 41: Do we need to simplify, or rely on the last snapping step?

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+        this.tiledGeometries.put(i, sliceIntoTiles(scaled, 0, 0.015625, i, extents.getForZoom(i)).getTileData());
```
bdon, line 42: We hardcode the buffer size here. We need a way to determine the actual buffer size taking into account any custom buffer size determination function.

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      JsonNode geoJson = mapper.readTree(bytes);
+      return new Clip(stats, GeoJSON.parseGeometry(geoJson));
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Clip GeoJSON is invalid");
+    }
+  }
+
+  // Copied from elsewhere in planetiler
```
bdon, line 68: we could make these public in planetiler-core?

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
```diff
@@ -0,0 +1,155 @@
...
+
+  @Override
+  public Map<String, List<VectorTile.Feature>> postProcessTile(TileCoord tileCoord,
+    Map<String, List<VectorTile.Feature>> map) throws GeometryException {
+    if (this.coverings.containsKey(tileCoord.z()) &&
+      this.coverings.get(tileCoord.z()).test(tileCoord.x(), tileCoord.y())) {
+      if (this.tiledGeometries.containsKey(tileCoord.z()) && this.tiledGeometries.get(tileCoord.z()).containsKey(tileCoord)) {
+        List<List<CoordinateSequence>> coords = tiledGeometries.get(tileCoord.z()).get(tileCoord);
+        var clipGeometry = reassemblePolygons(coords);
+        var clipGeometry2 = GeoUtils.fixPolygon(clipGeometry);
```
bdon, line 106: `fixPolygon` seems to work fine here instead of needing `snapAndFixPolygon`

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
wipfli: I would be surprised if we had to.

wipfli:
Do you store the clipping mask in world coordinates (web mercator x y of the 0/0/0 tile) or do you store it in tile coordinates?

wipfli:
I saw you have a covering tiles mechanism. I think you could get rid of this by using an STRtree https://locationtech.github.io/jts/javadoc/org/locationtech/jts/index/strtree/STRtree.html and use the tile bounds as an envelope for querying

tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java
bdon: use the default which is 4/256 pixels

bdon:
Results for small area in Monaco, expanding by 4/256 units: 

CleanShot 2025-02-03 at 20 23 31@2x

PR #387, opened 2025-02-21
Update Buildings.java: remove "else" so that polygon-buildings with an address get address data too

pietervdvn:
The current implementation is:

```
if( (sf.hasTag("building") && sf.canBePolygon()) || ... ){
  ... stuff ...
} else if (sf.hasTag("addr:housenumber")){
  ... actually add the housenumber ...
}
```

This means that a building (as polygon) will _never_ reach the else block where the address information is added. This means that the majority of buildings in Belgium (and many other countries) won't get address data included.

My fix is simply to remove this else. Note: I didn't test anything.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Buildings.java
@@ -97,7 +97,9 @@ public class Buildings implements ForwardingProfile.LayerPostProcesser {
       // Names should mostly just be for POIs
       // Sometimes building name and address are useful items, but only at zoom 17+
       //OsmNames.setOsmNames(feature, sf, 13);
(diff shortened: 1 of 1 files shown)


bdon:
Thanks, added unit test and bumped to v4.3.1 in https://github.com/protomaps/basemaps/pull/390

This should appear in the 2025-02-24 daily build and I'll update the hosted API when that appears.

pietervdvn:
Excellent! Thanks for your work. Protomaps is really becoming great to use!

bdon:
The hosted API is now the 20250224 build https://maps.protomaps.com/builds/ (b3 hash prefix `33682bf7`)

pietervdvn:
It works! 
image

PR #389, opened 2025-02-22
Use tag config lists instead of if-else blocks

wipfli:
This is a draft for restructuring the configuration of what happens when which tag is present.

The idea is to have a mapping from tag,value pair to feature property. Example:

```
highway=motorway -> minZoom = 3
highway=trunk -> minZoom = 6
```

This is done with `TagConfig` records which have a structure of tag, value, config_property_0, config_property_1 etc

I think with this approach we can be very explicit of what happens if this or that tag is present. Some of the exisiting logic is too complex for this scheme, and I added it instead still with some if-else after the list look-up block. 

Happy for some feedback before I continue with this and clean-up/test things...

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
@@ -15,132 +15,91 @@ import com.protomaps.basemap.locales.US;
 import com.protomaps.basemap.names.OsmNames;
 import java.util.*;
 
-public class Roads implements ForwardingProfile.LayerPostProcesser {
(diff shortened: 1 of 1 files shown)


bdon:
For the MultiExpression prototype I made (I think it is too complex + the wrong approach) there is a way to handle branching sub-logic for a tag match like here: https://github.com/protomaps/basemaps/commit/5265fdace3cf8d67fb5938b95727064c6a752e45#diff-93d007f0ae5256f93386fa849e597b81e1ee6bdbd59d13a9b74db70a96a5bca4R35 wonder if there is a way to extend TagConfig to be nested to do this?

wipfli:
Closing in favor of #399

PR #391, opened 2025-02-24
`--layer` cli config option

wipfli:
Adds a `--layer` cli config option which allows you to process only a single layer. For example, to do only the roads you can use
```
--layer=roads
```

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/Basemap.java
+++ b/tiles/src/main/java/com/protomaps/basemap/Basemap.java
@@ -29,7 +29,8 @@ import java.util.Map;
 
 public class Basemap extends ForwardingProfile {
 
-  public Basemap(NaturalEarthDb naturalEarthDb, QrankDb qrankDb, CountryCoder countryCoder, Clip clip, String renderOnly) {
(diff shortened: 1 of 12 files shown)


tiles/src/main/java/com/protomaps/basemap/Basemap.java
```diff
@@ -173,9 +194,29 @@ static void run(Arguments args) throws IOException {
           Paths.get(clipArg));
     }
 
+    List<String> renderOnlyOptions = List.of(
+      "Boundaries",
```
bdon, line 198: should these be lower case to match the vector tile names? maybe we could also retrieve them from the layer classes instead of hardcoding?

wipfli: I pull them from the classes now.

bdon:
what about calling it `--layers` or `--only-layers`? a little worried the "render" term is a bit over-applied...

wipfli:
It is only a single layer, so should we call it `--layer`?

bdon:
Yeah `--layer` also sounds good.

PR #392, opened 2025-02-24
Avoid variable shadowing in house number code

wipfli:
No functional change, just avoid shadowing the variable `feature` in the buildings layer.


wipfli:
Ok actually this the current code does not shadow the `feature` variable. But I still find it a bit easier to read this way...

PR #400, opened 2025-02-26
Update package-lock.json in styles

bdon:
* bump tsx and tsup versions


bdon:
Dependabot is on for these, however it can't automatically patch it because esbuild is on v0 so doesn't have safe SemVer upgrades.

PR #402, opened 2025-02-26
landuse unittests

wipfli:
Adds unit tests for the landuse `kind` selection logic.


bdon:
LGTM, there might be a way to make this less repetitive later with `@ParameterizedTest`

bdon (approved):

PR #406, opened 2025-03-03
roads unittests

wipfli:
Adds some unit tests for the roads layer. Not covered is:
* minZoomNames
* minZoomShieldText
* links
* tunnels

With the current test infrastructure I don't know how to easily test attributes that have a min zoom like 

```
.setAttrWithMinzoom("shield_text_length", shieldTextLength, minZoomShieldText)
```

One can supply a zoom to `assertFeatures` but then to check if the min zoom is correct one sort of needs to check the presence at minZoom and the absence at minZoom-1, but I don't know how to test for the absence of an attribute...

Anyway, this increases coverage already quite a bit.


wipfli:
CI passes now

bdon (approved):

PR #411, opened 2025-03-07
clean up water logic

wipfli:
When preparing water unittests I noticed that there are some things to clean up in the Water.java file. For example, if one checks for sf.hasTag("name") it is not necessary to later check again for sf.getTag("name") != null...

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Water.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Water.java
@@ -60,13 +60,7 @@ public class Water implements ForwardingProfile.LayerPostProcessor {
       }
 
       switch (sf.getString("featurecla")) {
-        case "Alkaline Lake" -> {
-          kind = "lake";
-        }
-        case "Lake" -> kind = "lake";
-        case "Reservoir" -> {
-          kind = "lake";
-        }
+        case "Alkaline Lake", "Lake", "Reservoir" -> kind = "lake";
         case "Playa" -> kind = "playa";
         case "Ocean" -> kind = "ocean";
       }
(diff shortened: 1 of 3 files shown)


tiles/src/main/java/com/protomaps/basemap/layers/Water.java
```diff
@@ -37,8 +37,6 @@ public void processPreparedOsm(SourceFeature ignoredSf, FeatureCollector feature
   public void processNe(SourceFeature sf, FeatureCollector features) {
     var sourceLayer = sf.getSourceLayer();
     var kind = "";
-    var alkaline = 0;
-    var reservoir = 0;
```
wipfli, line 41: Those are unused variables

wipfli:
CI seems to fail because it cannot find the geojson reader in planetiler. Strange...

```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/basemaps/basemaps/tiles/src/main/java/com/protomaps/basemap/feature/CountryCoder.java:[3,48] package com.onthegomap.planetiler.reader.geojson does not exist
Error:  /home/runner/work/basemaps/basemaps/tiles/src/main/java/com/protomaps/basemap/postprocess/Clip.java:[13,48] package com.onthegomap.planetiler.reader.geojson does not exist
```

wipfli:
Now it says

```
[INFO] Downloading from nexus-snapshots: https://s01.oss.sonatype.org/content/repositories/snapshots/com/onthegomap/planetiler/planetiler-core/0.8.4-SNAPSHOT/planetiler-core-0.8.4-SNAPSHOT.pom
Warning:  The POM for com.onthegomap.planetiler:planetiler-core:jar:0.8.4-SNAPSHOT is missing, no dependency information available
Warning:  The POM for com.onthegomap.planetiler:planetiler-core:jar:tests:0.8.4-SNAPSHOT is missing, no dependency information available
```

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
bdon: These properties are in tilezen, should we keep them for now and add them to the output vector tiles?

bdon: I guess we should treat this as cleanup for now, and also investigate how the Tangram styles ever used the alkaline/reservoir properties, because if they were never symbolized we don't need them...

wipfli: It is true though that alkaline was in the tilzen tiles, for ne see:

https://github.com/tilezen/vector-datasource/blob/89eeec6e98c6f5ae25d72970953ba93bb8b635a1/yaml/water.yaml#L250-L256

and for osm see:

https://github.com/tilezen/vector-datasource/blob/89eeec6e98c6f5ae25d72970953ba93bb8b635a1/yaml/water.yaml#L144-L147

wipfli: I suggest to remove the vector tile features properties `alkaline` and `reservoir`

wipfli: Removed also the OSM reservoir / alkaline

wipfli:
Since themeMinZoom is actually never used, there is another logic error which gets hidden by feature merge. The great lakes for example are always taken from the 10 m and the 50 m NE source layers. They appear only once in the final tiles because of feature merge post processing. But if we disable feature merge those appear twice:

image

wipfli:
Using themeMinZoom again now

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
bdon: no results for https://github.com/search?q=org%3Atangrams%20alkaline&type=code or reservoir=true in styles, so we can drop it

Issue #229, opened 2024-03-28
better minzoom ranking for water labels


bdon:
e.g. mediterranean should show up before all other sea labels

<img width="525" alt="Screenshot 2024-03-28 at 14 22 50" src="https://github.com/protomaps/basemaps/assets/77501/3be421ce-271c-4455-b9e3-ce0236f3f0d3">

nvkelso:
We should add the Natural Earth marine areas to the NaturalEarthDb config and then do same wikidata id join as in places layer, and default the sea min zoom to 7 or 8 for the fallback.

wipfli:
Closed by #416

PR #416, opened 2025-03-09
water matcher

wipfli:
Copied filter rules from tilezen's water.yaml file. Each copied filter rule was marked with "done". Each skipped item was marked with "skip" in https://github.com/wipfli/vector-datasource/pull/1

Skipped are the following

- NE coast lines
- intermittent
- boat
- reservoir
- alkaline

Rules in tilezen were evaluated top to bottom, we do pick the last. So I reversed the order of the filter rules, e.g., covered=yes appears last here, in tilezen it was first...

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Landcover.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Landcover.java
@@ -1,6 +1,7 @@
 package com.protomaps.basemap.layers;
 
 import com.onthegomap.planetiler.FeatureCollector;
+import com.onthegomap.planetiler.FeatureMerge;
 import com.onthegomap.planetiler.ForwardingProfile;
 import com.onthegomap.planetiler.VectorTile;
 import com.onthegomap.planetiler.geo.GeometryException;
@@ -8,6 +9,8 @@ import com.onthegomap.planetiler.reader.SourceFeature;
 import com.protomaps.basemap.feature.FeatureId;
 import java.util.List;
 import java.util.Map;
+
+import org.locationtech.jts.geom.Point;
 
 public class Landcover implements ForwardingProfile.LayerPostProcessor {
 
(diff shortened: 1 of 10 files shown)


wipfli:
Current planet tiles up to z15 for the water layer:

https://pmtiles.io/?url=https%3A%2F%2Fprotomaps.dev%2F%7Ewipfli%2Fwater-matcher.pmtiles#map=9.35/47.1402/8.4046

wipfli:
Since the water layer is closely related to the earth layer, I also touched the Earth.java file. Islands in the Mediterran for example appear now at the same zoom level as the corresponding hole appears in the water layer. At least mostly.

I found that if we use FeatureMerge.mergeNearbyPolygons without any buffer, i.e., buffer=0, then there are sometimes artifacts like spikes in earth and water polygons around zoom levels 6, 7, 8. I chose therefore to buffer but if the buffer is too large, e.g., buffer=0.5, then islands don't get corresponding holes in the water layer. The smallest buffer I saw had an effect on removing the spikes artifacts was the tile resolution, i.e., 0.0625.

Since I was touching the Earth.java file, I saw that the Antarctica fix is there but it should really be in the Landcover.java file. So I moved it over...

wipfli:
There are also some minor tweaks to the Landcover.java polygon merging. I can also revert those. I think if we want to go deeper into landcover I would do it in a separate pull request. Google maps buffers landcover by the way, I guess to avoid the gaps that you always get due to feature dropping and simplification:

image

wipfli:
Uploaded a new demo file to protomaps.dev with the layers water, earth, and landcover:

https://maps.protomaps.com/#map=2.08/-1.83/20.53&theme=light&lang=en&tiles=https://protomaps.dev/~wipfli/water-matcher.pmtiles&flavorName=light

I just noticed that the Atlantic Ocean label and similar ones are missing below zoom 6. Need to fix this...

wipfli:
Closes #229

wipfli:
Some place=sea come at z3, all others at z6. Looks like this for the less important sea labels at z6 in the Mediterranean:

image

wipfli:
Updated pmtiles file can be found at:

https://maps.protomaps.com/#map=5.32/22.263/45.628&theme=light&lang=en&tiles=https://protomaps.dev/~wipfli/water-matcher.pmtiles&flavorName=light

wipfli:
MinZoom for points should be 15, and not 12. Otherwise we get all these fountains etc on the map at z12

wipfli:
Also these bays from nodes in Croatia are a bit too much at z12. Better to show them only from z15 on... 
image
https://maps.protomaps.com/#map=12.36/44.37243/15.25533&theme=light&lang=en&tiles=https://protomaps.dev/~wipfli/water-matcher.pmtiles&flavorName=light

tiles/src/main/java/com/protomaps/basemap/layers/Water.java
```diff
@@ -21,262 +31,378 @@ public class Water implements ForwardingProfile.LayerPostProcessor {
...
-        kind = "water";
-        kindDetail = sf.getString("landuse");
-        reservoir = true;
-      } else if (sf.hasTag("leisure", "swimming_pool")) {
-        kind = "swimming_pool";
-      } else if (sf.hasTag("amenity", "swimming_pool")) {
-        kind = "swimming_pool";
-      }
+    if (!sf.tags().isEmpty() && sf.isPoint()) {
+      sf.setTag("_is_point", "yes");
```
bdon, line 330: should we move these internally into Matcher if we're going to use this feature in other layers?

tiles/src/main/java/com/protomaps/basemap/layers/Landcover.java
```diff
@@ -17,19 +19,42 @@ public class Landcover implements ForwardingProfile.LayerPostProcessor {
...
     String daylightClass = sf.getString("class");
     String kind = kindMapping.getOrDefault(daylightClass, daylightClass);
 
     // polygons are disjoint and non-overlapping, but order them in archive in consistent way
     Integer sortKey = sortKeyMapping.getOrDefault(daylightClass, 6);
 
-    features.polygon(this.name())
-      .setId(FeatureId.create(sf))
+    features.polygon(LAYER_NAME)
+      .setId(0)
```
bdon, line 52: We may actually want to set a separate ID depending on the landcover class (can do this in a follow up)

bdon (approved):

PR #422, opened 2025-03-14
Minzoom based styling

wipfli:
Although the tiles contain now nicely non-fragmented highways in the US, we still have a fragmented display at zoom 6 for example because the styles are not yet adjusted.

See Madison near Chicago for example: 
https://maps.protomaps.com/#flavorName=light&lang=en&map=6/43.263/-89.148

Explores how styling could be done with minzoom filtering, but actually I don't think this is the right approach. 

What we really should do is to demote the `kind` of non-interstate highways in the US from `kind=highway` to `kind=major_road`. Similarly, interstate `highway=trunk` should be promoted from `kind=major_road` to `kind=highway`.

### DIFF
--- a/styles/src/base_layers.ts
+++ b/styles/src/base_layers.ts
@@ -406,7 +406,10 @@ export function nolabels_layers(
         "all",
         ["!has", "is_tunnel"],
         ["!has", "is_bridge"],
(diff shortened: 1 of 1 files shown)


wipfli:
Closing in favor of #423

PR #423, opened 2025-03-14
Promote / demote US roads

wipfli:
Alternative solution to #422, fixes display fragmentation of roads around zoom 6 in the US.

Has not been tested yet. I will upload an archive later.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Roads.java
@@ -199,17 +199,20 @@ public class Roads implements ForwardingProfile.LayerPostProcessor, ForwardingPr
           trunk
           trunk_link
         """),
-      use("minZoom", 7)
+      use("minZoom", 7),
+      use("kind", "major_road")
     ),
     rule(
       with("_country", "US"),
       with("_r_network_US:US"),
-      use("minZoom", 6)
+      use("minZoom", 6),
+      use("kind", "major_road")
     ),
     rule(
       with("_country", "US"),
       with("_r_network_US:I"),
-      use("minZoom", 3)
+      use("minZoom", 3),
+      use("kind", "highway")
     )
   )).index();
 


bdon:
I think this solution is better than #422 since there are less vertices/features in the tiles.

nvkelso:
I'm curious to see before / after / comparisons between the approaches :)

> Alternative solution to https://github.com/protomaps/basemaps/pull/422, fixes display fragmentation of roads around zoom 6 in the US.

Is this because the style in the other PR needs a paired update?

> I think this solution is better than https://github.com/protomaps/basemaps/pull/422 since there are less vertices/features in the tiles.

What's removing the vertices / features? Feature merging elsewhere?

bdon:
Before: 
CleanShot 2025-03-17 at 11 57 28@2x

After:
CleanShot 2025-03-17 at 11 57 38@2x

nvkelso:
This smells like a styling issue where the `major_road`s are too faint compared with the `highway`s. 

Overzooming the images shows the same geometry vertexes are included in each before and after. The after draws them with a thinner styling.

Going with this this data hack approach risks either creating flicker on the next zooms in when the real raw classification is shown, or perpetuates the reclassification giving false impression of the road classes (sometimes ok if just for early zooms). Spice to taste.

wipfli:
Here is a demo of the roads layer with the changes of this pull request:

https://maps.protomaps.com/#map=7.29/41.498/-88.041&theme=light&lang=en&tiles=https://protomaps.dev/~wipfli/promote-demote-us-roads.pmtiles&flavorName=light

wipfli:
I think styling the roads based on the `kind` attribute is the right thing to do. It represents the kind of road. Important ones show up earlier.
The `highway` tag in OpenStreetMap on the other hand represents more the class of road in the national network and not necessarily the importance of the road. For example in the US, if a highway=primary is part of the US:US network, then that road is more important than a highway=primary which is not part of any network. So the highway tag alone does not contain the full road importance...

Issue #294, opened 2024-09-05
Add Island Names


rami-dv:
Islands are important and often notable map features, and island labels should be included in the tiles.

With the island's geometry (if it's a multipolygon relation), it should be possible to infer a reasonable minzoom for the label point, such as done in the [planetiler openmaptiles profile](https://github.com/openmaptiles/planetiler-openmaptiles/blob/5673b877afea62be787dba899c38f65d9d812630/src/main/java/org/openmaptiles/layers/Place.java#L285-L312). In that profile, the multipolygons are set with a minzoom depending on their area, while all label points are added at zoom 12.

Currently, island name points are the only data that I've needed, other than a protomaps extract, in order to make a solid basemap. The logic used in the OMT profile seems solid to me, and I think the island label points would fit in just fine in the `physical_point` layer.

nvkelso:
Great idea, can you assemble a PR?

bdon:
The OMT code is licensed differently from this repo and not license-compatible, so we can't derive any code from that.

Otherwise it should just be as simple as deriving a rank from the area? @nvkelso since we are getting rid of `physical_point`, which Tilezen layer would islands go in?

nvkelso:
There isn't a license issue referring to Tilezen (@bdon rightly points out the OMT issue). 

Tilezen does much the same thing but varies the zoom a bit and does archipelago, island, and islet here:

https://github.com/tilezen/vector-datasource/blob/master/yaml/earth.yaml

nvkelso:
Walkabout is most relevant, fully featured style:

https://tangrams.github.io/walkabout-style/#15/37.7926/-122.4003

wipfli:
Note sure how relevant it is for the discussion here, but I use QRank for island labels on https://wipfli.github.io/swiss-map/#map=8.61/42.4284/10.2657

https://github.com/wipfli/swiss-map/blob/e9f69c8088e1d7ee307bae6c35904ca20933145b/planetiler/src/main/java/com/onthegomap/planetiler/examples/QRank.java#L123-L126

There are issues though. For example the island relation for Sardigna does not point to the wikidata object of the main island, which has high qrank, but rather to the wikidata item of the collection of islands of Sardigna and the small neighbors which has low qrank...

bdon:
nice San Juan islands example from bubble-wrap: https://tangrams.github.io/bubble-wrap/#10.029166666666685/48.5077/-122.6431

[...]

PR #424, opened 2025-03-15
add island names [#294]

bdon:


tiles/src/main/java/com/protomaps/basemap/layers/Earth.java
```diff
@@ -64,6 +64,15 @@ public void processOsm(SourceFeature sf, FeatureCollector features) {
 
       OsmNames.setOsmNames(feat, sf, 0);
     }
+
+    if (sf.canBePolygon() && sf.hasTag("place", "island")) {
+      var feat = features.innermostPoint(LAYER_NAME)
+        .setId(FeatureId.create(sf))
+        .setAttr("kind", "island")
+        .setMinPixelSize(10)
```
wipfli, line 72: It is actually a good idea to use a min pixel size in pixels. In the water layer we compute water body names based on zoom, but I think the logic should be the same as here, i.e., just a constant pixel value...

wipfli:
Tuned the parameters a little, now it looks like this:

image

wipfli (approved):

PR #425, opened 2025-03-17
Special post-processing for rivers

wipfli:
After merging the water matcher pull request, we found that some rivers got dropped because we increased the minimum polygon area to 10 pixels2.

Here we apply special casing for rivers in the post-processing step, such that river polygons are kept down to 1 pixel2 area.

However, not that sometimes OSM features don't tag stuff as rivers. So in Bogota for example there will still be feature dropping because the merging fails (kind_detail=river is absent). Example: https://www.openstreetmap.org/relation/12757766#map=11/5.6744/-74.4660


bdon:
I think #428 can solve this too

wipfli:
Closing in favor of #428

PR #427, opened 2025-03-20
pass through regular font override to get_name_block

Shane98c:
Noticed that my font overrides weren't making it through in all cases, which caused failed requests for `Noto Sans Regular`. This change makes sure to pass through `regular` to the `get_name_block()` function.


bdon:
5.0.1 published, thanks!

PR #428, opened 2025-03-21
set min earth size back to 1px

bdon:
CleanShot 2025-03-21 at 13 13 43@2x

make sure taiwan and most of the philippine archipelago shows up at zoom=0 instead of disappearing. 

The planet build size is 120.3 GB


wipfli (approved):
That's also a good solution

PR #431, opened 2025-03-23
Export get_multiline_name and get_country_name

russss:
I've had to fork the style to customise it, but it would be nicer to import these functions rather than copy them.


wipfli (approved):

bdon:
Thanks. I just published this as 5.0.2 (mistake, it should have been 5.1.0 but I'll live with it)

PR #434, opened 2025-03-25
Places matcher

wipfli:
Moves Places.java to the matcher syntax. I left the logic as-is.

You can compare before-after here:

https://wipfli.github.io/tiles-comparator-protomaps/?left=https://protomaps.dev/~wipfli/places-main.pmtiles&right=https://protomaps.dev/~wipfli/places.pmtiles#map=9.51/47.3935/8.4797

That's a small tool that I vibe-coded to show a left and a right pmtiles file with the same style.json.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Tiles 4.8.2
+------
+- port places to use Matchers [#434]
+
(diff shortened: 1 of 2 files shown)


bdon:
Can you also bump the version in `Basemap.java`? I think this can be a patch version

bdon (approved):

PR #437, opened 2025-03-27
places grid

wipfli:
Copies the pre-collide changes from #300 over to latest main.

### DIFF
--- a/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
+++ b/tiles/src/main/java/com/protomaps/basemap/layers/Places.java
@@ -227,27 +227,12 @@ public class Places implements ForwardingProfile.LayerPostProcessor {
 
   private static final ZoomFunction<Number> LOCALITY_GRID_SIZE_ZOOM_FUNCTION =
     ZoomFunction.fromMaxZoomThresholds(Map.of(
-      3, 24,
-      4, 24,
-      5, 24,
-      7, 24,
-      8, 24,
-      9, 24,
-      10, 24,
-      11, 24,
       14, 24,
       15, 16
     ), 0);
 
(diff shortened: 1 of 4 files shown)


tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -210,15 +212,30 @@ static int getSortKey(double minZoom, int kindRank, int populationRank, long pop
...
   private static final ZoomFunction<Number> LOCALITY_GRID_SIZE_ZOOM_FUNCTION =
     ZoomFunction.fromMaxZoomThresholds(Map.of(
-      6, 32,
-      7, 64
+      3, 24,
+      4, 24,
+      5, 24,
+      7, 24,
+      8, 32,
+      9, 32,
```
wipfli, line 220: I have noticed that some features are present in zoom 7, then they disappear at zoom 8, and reappear at zoom 9:

Example Toulon

Zoom 7 many points in orange circles:

image

Zoom 8 only one point for Toulon:

image


Zoom 9 many points reappear:

image

wipfli: Maybe this is due to the increase in the label grid cell size. Larger cell means more features fall into the same cell means fewer are retained...

wipfli:
Here is a jsbin showing the difference to main. Orange disks are from main, and red points are from this branch:

https://jsbin.com/libevelixe/1/edit?html,output

image

wipfli:
Here is a side-by-side comparision: https://wipfli.github.io/tiles-comparator-protomaps/?left=https://protomaps.dev/~wipfli/places-main.pmtiles&right=https://protomaps.dev/~wipfli/places-437.pmtiles

PR #440, opened 2025-03-28
Make the style work with older tile builds missing sort_rank [#436]

bdon:
@wipfli without this we get console warnings `Expected value to be of type number, but found null instead.` if the style depends on `sort_rank` but the tileset is a 4.x that doesn't contain it.

If we didn't want to do this then we would require a SemVer bump for this kind of change which seems painful.


wipfli (approved):

wipfli:
Good catch!

PR #441, opened 2025-03-28
region labels

wipfli:
Restricts region labels (place=state or province) to the following countries:

- USA
- Brazil
- Canada
- India
- China
- Australia

Those are large countries where one region alone can be larger than your typical European country.

On the style side, it move the region layer below the locality layer. This is needed for example in the US East coast becuase US state labels otherwise collide away the big city names like New York or Washington.

Fixes also abbreviated names below zoom 6. Before it used `name:short` but actually that key is not in the tiles anymore. So now it uses `ref:en` if available and `ref` otherwise. This makes sure we get a Latin abbreviation in China.

A PMTiles file of the places layer is available at https://protomaps.dev/~wipfli/region-labels.pmtiles


bdon (approved):

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
```diff
@@ -222,10 +237,19 @@ static int getSortKey(double minZoom, int kindRank, int populationRank, long pop
     ), 0);
 
   public void processOsm(SourceFeature sf, FeatureCollector features) {
-    if (!sf.isPoint() || !sf.hasTag("name")) {
+    if (!sf.isPoint() || !sf.hasTag("name") || !sf.hasTag("place")) {
       return;
     }
 
+    try {
+      Optional<String> code = countryCoder.getCountryCode(sf.latLonGeometry());
```
bdon, line 245: Good use of country coder, though I wonder if it is going to be accurate for all regions especially if they have "odd" geometries like exclaves not fully contained? Anyways, this is better than nothing for now

bdon:
LGTM, can you fix the conflict and bump the minor versions on styles/tiles?

tiles/src/main/java/com/protomaps/basemap/layers/Places.java
wipfli: Not fully sure here either. There are also different country codes for example for France (Europe+overseas) and Metropolitan France (Europe only), so one has to be careful with this...

PR #445, opened 2025-04-02
Use tilezen country minzooms

wipfli:
Ports the country minzooms over from tilezen, see https://github.com/tilezen/vector-datasource/blob/master/spreadsheets/min_zoom/country.yaml

Uses `minZoomProtomaps = round(minZoomTilzen) - 1`

Removes conflation with Natural Earth for country minzoom querying.

This should be convenient in the future to tune the minzoom of places like Vatican City or San Marino which currently collide away the city labels of Rome and Fierenze.

Live demo: https://maps.protomaps.com/#flavorName=light&lang=en&map=3.5/40.46/10.9&tiles=https://protomaps.dev/~wipfli/country-minzoom.pmtiles

Merge only after #437 

The conflation between Wikidata ID and OpenStreetMap and Tilezen was done with this python script: https://gist.github.com/wipfli/3f6546e107b7cac214054e9c6b8d92ee


wipfli:
Switched to a hash map to have a more compact notation. Joins based on Wikidata ID now.

bdon (approved):

PR #448, opened 2025-04-05
import only types from @maplibre/maplibre-gl-style-spec

bdon:
@wipfli I am trying to make it so that projects depending on `@protomaps/basemaps` do not need to bring in maplibre-style-spec as a dependency


wipfli (approved):
Those look good. Would be nice if the linter could do this sort of optimization

PR #451, opened 2025-04-08
Use text-variable-anchor

wipfli:
MapLibre GL can place text labels such that the anchor, i.e., the icon/symbolizer location, is to the left, right, top, or bottom of the label. So far we are only using labels where the anchor is to the left.

This pull request proposes to have a variable anchor, with bottom taking precedence over left, and right. As a result, more labels can be shown which is convenient in dense areas.

Demo: https://wipfli.github.io/debug-assets/text-variable-anchor/index.html#map=1.99/0/0

### DIFF
--- a/styles/src/base_layers.ts
+++ b/styles/src/base_layers.ts
@@ -1878,11 +1878,11 @@ export function labels_layers(
         "text-variable-anchor": [
           "step",
           ["zoom"],
-          ["literal", ["bottom-left", "top-left", "bottom-right", "top-right"]],
+          ["literal", ["bottom", "left", "right", "top"]],
           8,
           ["literal", ["center"]],
         ],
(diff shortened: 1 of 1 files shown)


wipfli:
image
image
image

nvkelso:
Nice!

nvkelso (approved):

bdon:
I feel like it looks wrong for a townspot to be positioned diagonally next to stacked text. That might mean we should limit variable anchor to only the left or right, so the townspot is equidistant to both languages.  @nvkelso thoughts?

CleanShot 2025-04-09 at 16 33 58@2x

wipfli:
Using default anchor bottom
image

wipfli:
Adding a render test for this in #485

PR #466, opened 2025-04-23
sonar string literals

wipfli:
In sonar cloud we have maybe around 100 issues about duplicated string literals, e.g., the string "highway" appears 33 times in the file Roads.java.

We can suppress this warning, but actually it is probably better to store these strings in a central location and reference them. That avoids typos.

This is a draft so far.


wipfli:
This is too complicated. Closing in favor of #468

PR #472, opened 2025-04-25
refactor: grayscale style for better contrast and flatter colors

0PandaDEV:


wipfli:
Thanks! Can you share some before / after screenshots?

0PandaDEV:
Sure, I tested it on basemaps flavours

## Before

image

## After

image

0PandaDEV:
@wipfli Yes, but then I could argue that the current grayscale is more like a light mode instead of gray :)

bdon:
This is changing one of the default flavors to a fundamentally different appearance.

The https://github.com/protomaps/basemaps-flavors is intended as a namespace for third-party contributed themes. It isn't built yet, but is intended to have open contributions as long as you 1) provide a flavor that is sufficiently different from existing ones and 2) give your flavor a unique name. Does that better meet your goals for this contribution?

bdon:
Hi @0PandaDEV 

with `@protomaps/basemaps` v5.4.0 you can now do this:

```
npm install @protomaps/basemaps
npx generate_style out.json https://example.com/tiles.json flat.ts en
```

That will output the full style to `out.json`

`flat` can also be a JS ES6 module or a JSON

Let me know if that works for your use case.

PR #473, opened 2025-04-28
chore: improved Dockerfile to utilize cached layers

mxzinke:
It was kind of annoying to me, doing a complete refetch of the dependencies on each build within docker.

### DIFF
--- a/tiles/Dockerfile
+++ b/tiles/Dockerfile
@@ -1,6 +1,21 @@
-FROM maven:3-eclipse-temurin-22-alpine
+# Stage 1: Build
+FROM maven:3-eclipse-temurin-22-alpine AS build
 WORKDIR /tiles
-COPY src src
+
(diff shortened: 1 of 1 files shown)


wipfli:
@bdon do we actually want to keep maintaining this dockerfile?

wipfli:
Thanks for opening the pull request in any case @mxzinke. I just never used that file before and maybe we could abandon docker for the build altogether...

bdon (approved):

bdon:
Thanks, this is a nice improvement but it still looks like it contacts the repository at `mvn clean package` step without the `-o` flag? (I couldn't get that to work)

I think we should have a Dockerfile for the audience that prefers that, though to make this a good experience we would also need to provide some instructions around mounting the `data` directory volume from the host so it doesn't re-download sources every time.

PR #477, opened 2025-04-30
Add Monaco build test

wipfli:


tiles/monaco.sh
```diff
@@ -0,0 +1,6 @@
+java -Xmx2g -jar target/protomaps-basemap-HEAD-with-deps.jar \
+    --osm_url=https://protomaps.dev/~wipfli/fixtures/monaco-2025-02-15.osm.pbf \
```
bdon, line 2: If this is small enough can we commit it to the repo to make this more self-contained?

If not, maybe the r2-public bucket

wipfli: The file is 128 KB. I am at the moment looking into render tests as well and we might get more pmtiles and expected images. Let's think of this pull request for now as draft...

wipfli: Committed it to git now

wipfli:
This now looks all ready:

image

I made a change in the styling to have a failing render test above. You can click on the render test ci and then go to the "Upload render tests results.html on failure" job. This brings you to a link where you can download a zip file containing the CI's render test results.html file. Download an open it and you will see this:

image

render-tests/tests/roads/default/style.json
```diff
@@ -0,0 +1,22 @@
...
+      "height": 256,
+      "flavor": "light",
+      "lang": "en"
+    }
+  },
+  "center": [
+    7.420737,
+    43.736756
+  ],
+  "zoom": 15,
```
wipfli, line 15: Here we define where the test is centered

render-tests/generate_pmtiles.sh
```diff
@@ -0,0 +1,9 @@
+mkdir -p pmtiles
+cd ../tiles
+mvn clean package -DskipTests
+java -Xmx2g -jar target/protomaps-basemap-HEAD-with-deps.jar \
+    --osm_path=../render-tests/fixtures/monaco-2025-02-15.osm.pbf \
```
bdon, line 5: I think in the future we can use SliceOSM to generate a more linguistically diverse, yet compact OSM extract and then render the PMTiles off of that

bdon (approved):

bdon:
LGTM, can merge after reverting the intentional breakage for CI failure

PR #485, opened 2025-05-23
Update render tests

wipfli:
Adds render tests for different recent issues and pull requests. See README.md for documentation.

Todo:

- [x] multilingual labels
- [x] Devanagari


wipfli:
In the runner it only takes like 20 seconds to download land and water polygons:

image

bdon:
It seems a little annoying to hammer that OSM.de server in our CI, is there a way to cache it in GitHub actions?

render-tests/generate_pmtiles.sh
```diff
@@ -1,9 +1,38 @@
+#!/bin/bash
+set -euo pipefail
+
+target_subdir="${1:-}"
+
+if [[ -n "$target_subdir" ]]; then
+  echo "Processing only subfolder: $target_subdir"
+  mapfile -t input_files < <(find "$target_subdir" -type f -name "input.osm.pbf")
```
bdon, line 8: `mapfile` doesn't exist on macOS, is there an alternative?

wipfli: Also, if you think this file is too much of a mess or is hard to maintain, I am happy to write the functionality in typescript instead...

bdon:
Tests are working for me! LGTM

One comment, if I don't have the pmtiles prepared the error message from running the tests isn't helpful, it just reports the test timed out - is there a good resolution for that?

bdon (approved):

PR #488, opened 2025-05-27
Add distinc symbolizer for capital cities

wipfli:


wipfli:
The old threshold was actually too high to capture some changes. I think the pixelmatch library gives different weights to pixel differences based on perceived color changes. Since the green and gray tones are all close together, we can use quite a low threshold.

PR #490, opened 2025-05-27
Fix natural parks

wipfli:


styles/src/base_layers.ts
```diff
@@ -66,19 +66,15 @@ export function nolabels_layers(
...
-        "nature_reserve",
-        "forest",
         "golf_course",
         "wood",
-        "nature_reserve",
         "forest",
         "scrub",
         "grassland",
         "grass",
+        "glacier",
```
wipfli, line 77: We forgot this one in the past. Now high-zoom glaciers are visible again. Maybe I should add a render test for this...

bdon: Yes, I think the render test is most useful for this class of issue, but do we have enough data in the CI sample data?

jgillich:
:+1: 

FWIW this issue extends far beyond Amsterdam, the whole Swabian Jura is rendered as a park, even cities. Before and after this change:
<p float="left">
Screenshot From 2025-08-19 19-20-00Screenshot From 2025-08-19 19-17-18
</p>

Issue #334, opened 2024-11-28
Exclude roads with `access=no` tag


wipfli:
**Location**
https://maps.protomaps.com/#map=12.64/44.88566/10.25981&theme=light&lang=en&tiles=https://demo-bucket.protomaps.com/v4.pmtiles

https://www.openstreetmap.org/way/891239306#map=14/44.88768/10.26492

https://wiki.openstreetmap.org/wiki/Tag:access%3Dno

**Screenshots**

Protomaps show that road like all other highway=motorway

![image](https://github.com/user-attachments/assets/1f97e415-3d11-4c4e-92df-f7ff65c61068)

Google maps excludes it completely

![image](https://github.com/user-attachments/assets/6a319777-809d-4a85-9b7c-7aca950ec853)

Apple maps gives it a weaker color and higher minzoom than highway=motorway
![image](https://github.com/user-attachments/assets/424bc98f-ca9f-4ebe-8bc8-0e729fcb2d40)

bdon:
What about the same area on http://tangrams.github.io/bubble-wrap/ and https://demo.tangram.city ?

nvkelso:
Roads like these should only be visible at max zooms. But they are useful to see.On Dec 8, 2024, at 19:59, Brandon Liu ***@***.***> wrote:﻿
What about the same area on http://tangrams.github.io/bubble-wrap/ and https://demo.tangram.city ?

wipfli:
[Bubble Wrap 
](https://tangrams.github.io/bubble-wrap/#12/44.8857/10.2599)![image](https://github.com/user-attachments/assets/fae8439b-27d7-4384-97b9-ed3322e8ca15)

[Tangram City
](https://demo.tangram.city/#12.64/44.885736191733116/10.259771942567971)![image](https://github.com/user-attachments/assets/f1878db3-b113-4050-bd1b-ce35f88a3948)

bdon:
@wipfli  I think we should make the zoom appear behavior similar to Tilezen, also considering also the 1 level in zoom difference between mapzen and leaflet.

wipfli:
Can be implemented with matcher rules #407

PR #491, opened 2025-05-28
Render access=no,private roads only at zoom 15

wipfli:
Fixes https://github.com/protomaps/basemaps/issues/334


bdon:
Seems like the check is working now! LGTM after version bump

PR #495, opened 2025-06-03
Remove building IDs and building minzoom 15

wipfli:
Removes the IDs from buildings (sets all to a value of 1), pushes the building minzoom to 15, removes the sort_rank tag on buildings. 

Planet buildings only is on main 19 GB (100%) and here it is 12 GB (63%).

The computation time on the buildings layer only went from 40 min (100%) on main to 21 min (53%) on this branch.


bdon:
We definitely want generalized buildings though, that's one of the most common requests is to show buildings at earlier zooms instead of deferring them to z13+ or higher. 

I don't believe sort_rank impacts the total size of the archive much, after compression, because they are all the same value?

mxzinke:
Dont we need the sort rank for overlapping buildings (usage in 3D maps)?

bdon:
Sort rank exists for legacy purposes (tilezen) and isn't strictly needed for building parts in MapLibre.

In general I think the building generalizations should stay as they are and the IDs are necessary for feature-picking like in the pmtiles.io inspector.

Issue #499, opened 2025-06-04
styles: directional arrows for roads


prusswan:
**Description**
Just checking if there are any plans to introduce arrows or other road markings

This would bring it closer to parity with default styles from OpenMapTiles, Versatiles, OpenStreetMap etc

bdon:
See https://github.com/protomaps/basemaps/pull/501

bdon:
Released in styles 5.5.0

PR #501, opened 2025-06-05
style: show oneway road arrows [#499]

bdon:


wipfli:
The arrows should have a tail - something like this:

image

On my not-so-great laptop display it is otherwise too hard to see the direction with the current sprite

image

bdon:
we should make the arrow bigger at all zooms, but that would also require changing the road widths to be sufficiently wide

see http://tangrams.github.io/bubble-wrap/#18.20833333333332/37.79213/-122.40090 for comparison

If the arrows are that size do you think we still need the tail?

wipfli:
I think scaling normal raster png sprites is bad. I made the size of the original arrow now 0.5 times smaller and for me it looks OK. Let me know what you think.

Lake area in Zürich

image

bdon:
Looks good!

wipfli:
CI is green now @bdon

PR #507, opened 2025-08-13
Reduce minZoom for subway

spencermountain:
hey there Brandon et al - I'm a huge fan of the project.

I'd like to propose reducing the `minZoom`  for the subway road-layer data. My motivation is to see a city's subway network from the zoom level of a normal city. This layer is currently working great, but only at a very-close zoom.

Please let me know if this change is appropriate, or can be improved. I know there are a lot of plans for future transit data. This subway data is not shown in any of the default basemap flavours, so may have a minimal impact.
cheers


jgillich:
(apologies, I linked the wrong issue)

PR #518, opened 2025-10-24
Shields

bdon:
Implementation of region-aware highway shields


tiles/src/main/java/com/protomaps/basemap/locales/US.java
```diff
@@ -19,10 +19,10 @@ public Shield getShield(SourceFeature sf) {
       String shieldText = firstRef;
       if (firstRef.startsWith("US ")) {
         shieldText = firstRef.replace("US ", "");
-        network = "US:US";
+        network = "US";
```
nvkelso: Why are you changing this from `US:US` to `US`? The pattern we've generally used is:

> `{country_code}:{network}`

bdon: Good catch, fixed in https://github.com/protomaps/basemaps/pull/606

PR #520, opened 2025-11-02
Update base_layers.ts: add 3D

pietervdvn:
Hey,

You've got an awesome project, which apparently also included 3D-data. I wasn't aware of that as the basestyles don't leverage it.

I've added a small snippet to the basestyle to use that 3D-support to show it of. Copied it out of my own fork of the style (MapComplete-Sunny), so it _should_ work, but I have to admit that I didn't test it in the environment here.

Anyway, here is how it looks in [MapComplete](https://dev.mapcomplete.org/cyclofix.html?z=16.6&lat=51.054371466030716&lon=3.7246584907155693): 

image

### DIFF
--- a/styles/src/base_layers.ts
+++ b/styles/src/base_layers.ts
@@ -601,34 +601,22 @@ export function nolabels_layers(
       },
     },
     {
-      "id": "3d-buildings",
-      "source": "protomaps",
+      id: "3d-buildings",
+      source: "protomaps",
       "source-layer": "buildings",
-      "type": "fill-extrusion",
-      "paint": {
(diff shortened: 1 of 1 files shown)


bdon:
My current plan is to make 3D buildings a [Flavor](https://docs.protomaps.com/basemaps/flavors#overriding-defaults) parameter, boolean true/false. The majority of basemap use cases are 2D only. the styles hosted at https://protomaps.com/api are pre-generated style JSONs where you have no control over Flavor parameters, but loading from the npm package `@protomaps/basemaps` lets you pass in type-safe parameter overrides.

PR #571, opened 2026-03-03
add DuckDB script for downloading monaco or a custom area from Overtu…

bdon:
…re as Parquet file [#541]

@migurski @wipfli 

This seems more minimal than demanding installation of https://docs.overturemaps.org/getting-data/overturemaps-py/ or another custom tool?

### DIFF
--- a/tiles/download-overture.sh
+++ b/tiles/download-overture.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+REL="${1:-2026-02-18.0}"
+BBOX="${2:-7.408446,43.722901,7.4405,43.752481}"
+OUT="data/sources/${3:-monaco.parquet}"
+
+IFS=',' read -r XMIN YMIN XMAX YMAX <<< "$BBOX"
+
+echo "Downloading release $REL: $BBOX to $OUT"
+
+duckdb -c "
+COPY (
+  SELECT *
+  FROM read_parquet(
+    's3://overturemaps-us-west-2/release/${REL}/**/*.parquet',
+    hive_partitioning=1, filename=1, union_by_name=1
+  )
+  WHERE theme IN ('transportation','places','base','buildings','divisions')
+    AND bbox.xmin <= ${XMAX}
+    AND bbox.xmax >= ${XMIN}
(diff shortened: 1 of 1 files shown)


wipfli:
I think adding such a script is a good idea. I get the following error:

```
Downloading release 2026-02-18.0: 7.408446,43.722901,7.4405,43.752481 to data/sources/monaco.parquet
HTTP Error:
HTTP GET error on '/?encoding-type=url&list-type=2&prefix=release%2F2026-02-18.0%2F' (HTTP 400)
Wrote data/sources/monaco.parquet
```

wipfli:
```
➜  overture git:(main) ✗ duckdb --version
v1.2.1 8e52ec4395
```

wipfli:
I am not good at reading bash. I would prefer if we could hard-code the Overture release and the bounding box in the SQL. In the current state I would need to ask a chatbot or read about bash's syntax to know how to use this script.

bdon:
What if we added help text?

```
./download-overture.sh --help
Usage: ./download-overture.sh [RELEASE] [BBOX] [OUTPUT]

Downloads Overture Maps data for a bounding box as a Parquet file.

Arguments:
  RELEASE  Overture release version              (default: 2026-02-18.0)
  BBOX     xmin,ymin,xmax,ymax in WGS84         (default: 7.408446,43.722901,7.4405,43.752481)
  OUTPUT   Output file path                      (default: data/sources/monaco.parquet)

Example:
  ./download-overture.sh 2026-02-18.0 -122.52,37.70,-122.35,37.83 data/sources/sf.parquet
```

migurski:
The help text really improves it. They expire Overture releases quite quickly btw, I think keeping just two around at any given point. My [Python predecessor to this script](https://gist.github.com/migurski/8765492ef831457d7c21c20f4e454e1b#file-get-overture-py) already does not work.

> ➜  overture git:(main) ✗ duckdb --version
> v1.2.1 8e52ec4395

Do you still see this error if you upgrade to a more recent DuckDB? Version 1.4.x should work, and I recall seeing exactly your error when I was previously stuck at version 1.1.x.

bdon:
* added help text
* use DuckDB to query for the latest Overture release from STAC instead of hardcoding

PR #591, opened 2026-04-08
Overture source: allow hive-partitioned directory input

bdon:
@migurski


migurski (approved):
No notes, looks great, excited to be able to use this!

PR #594, opened 2026-04-15
Fix link to app

gtitov:
Github reads maps.protomaps.com as a relative link and tries to open https://github.com/protomaps/basemaps/blob/main/maps.protomaps.com


bdon (approved):

bdon:
Thanks!

PR #624, opened 2026-07-20
add skills

bdon:
@migurski @wipfli would like your thoughts on the way to structure this for letting an agent drive the entire cycle:

* download a extract from SliceOSM
* build the tiles
* do programmatic, and later visual, analysis of the built output

### DIFF
--- a/.claude/skills/analyze-tiles/SKILL.md
+++ b/.claude/skills/analyze-tiles/SKILL.md
@@ -1,0 +1,123 @@
+---
+name: analyze-tiles
+description: >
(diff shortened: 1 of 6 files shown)


wipfli:
This look like a good addition, thanks for creating these skills.
Two points that come to mind both in the overture skill:
1. The overture types have changed over time. Maybe they have fully stabilized now but if new types appear we should update this skill.
2. Maybe instead of going into a .venv folder to get to the overture cli tool suggest using `uvx overturemaps download ...`

bdon:
I updated the Overture skill to use DuckDB and the included `download-overture.sh`.

PR #635, opened 2026-09-10
app: upgrade to maplibre-gl 6.9

bdon:
* obviates need for loading rtl text plugin


wipfli:
We could now probably also remove the positioned glyph fonts. Do you have any plans in that direction?

bdon:
> We could now probably also remove the positioned glyph fonts. Do you have any plans in that direction?

it's not clear to me if 6.9 depends on an empty `glyphs` key to accomplish this? if so we can make a breaking change to remove positioned glyphs, RTL and font PBFs

